PR #5, opened 2019-09-20
Removed duplicate reference resolves

sunspikes:
References are already resolved in the `Reader`,

https://github.com/cebe/php-openapi/blob/master/src/Reader.php#L70
https://github.com/cebe/php-openapi/blob/master/src/Reader.php#L96

### DIFF
--- a/src/PSR7/SchemaFactory/JsonFileFactory.php
+++ b/src/PSR7/SchemaFactory/JsonFileFactory.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace OpenAPIValidation\PSR7\SchemaFactory;
 
 use cebe\openapi\Reader;
-use cebe\openapi\ReferenceContext;
 use cebe\openapi\spec\OpenApi;
 use function realpath;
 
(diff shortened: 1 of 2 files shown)


cebe:
> References are already resolved in the `Reader`,

while the reader does resolve references, it does not do so recursively.

There is currently no built-in way to resolve references recursively until all levels are covered.

See https://github.com/cebe/php-openapi#reading-api-description-files-and-resolving-references , there is a note about this:

> Note: Resolving references currently does not deal with references in referenced files, you have to call it multiple times to resolve these.

I have some open issues about references and I will think about how this can be solved. Until then I think this code should not be removed.

sunspikes:
@cebe I see, thank you for the clarification. In my application i have a huge openapi file with multiple level of schema definition and the openapi validator crashes (even if i set the memory_limit=-1, btw it doesn't use much memory) because of this. If i comment these lines out everything works.

scaytrase:
Well, the best thing I can suggest is to make this calls optional, enabled by default, so one could disable resolution for single-file specs while creating schema object

scaytrase:
I think we can close this since we decided to fix this in dependency

PR #6, opened 2019-09-20
Fix UniqueItems validator

sunspikes:
This fixes the `UniqueItems` validator behavior


lezhnev74:
awesome!

sunspikes:
@lezhnev74 thank you!

PR #7, opened 2019-09-20
Callback validation

### LINKED ISSUE #3: Validation of callbacks
Some way to validate the request and response of callbacks would be great.

I think the validation per se would work, but there needs to be some other way to resolve the specs. Right now, the specs are loaded based on a passed OperationAddress which doesn't fit for callbacks.

### Adding more docs:

https://swagger.io/docs/specification/callbacks/
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#callback-object

digilist:
Reopening my PR to implement callback validation from the [previous repository](https://github.com/lezhnev74/openapi-psr7-validator/pull/69) here again.

fixes #3

### DIFF
--- a/src/PSR7/CallbackRequestValidator.php
+++ b/src/PSR7/CallbackRequestValidator.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\OpenAPIValidation\PSR7;
+
+use cebe\openapi\spec\OpenApi;
+use League\OpenAPIValidation\PSR7\Exception\ValidationFailed;
+use League\OpenAPIValidation\PSR7\Validators\BodyValidator\BodyValidator;
+use League\OpenAPIValidation\PSR7\Validators\CookiesValidator\CookiesValidator;
+use League\OpenAPIValidation\PSR7\Validators\HeadersValidator;
+use League\OpenAPIValidation\PSR7\Validators\QueryArgumentsValidator;
+use League\OpenAPIValidation\PSR7\Validators\ValidatorChain;
+use Psr\Http\Message\RequestInterface;
+
+final class CallbackRequestValidator implements ReusableSchema
+{
+    /** @var OpenApi */
+    protected $openApi;
+    /** @var MessageValidator */
+    protected $validator;
+
+    public function __construct(OpenApi $schema)
+    {
+        $this->openApi   = $schema;
+        $finder          = new SpecFinder($this->openApi);
+        $this->validator = new ValidatorChain(
+            new HeadersValidator($finder),
+            new CookiesValidator($finder),
+            new BodyValidator($finder),
+            new QueryArgumentsValidator($finder)
+        );
+    }
+
+    public function getSchema() : OpenApi
+    {
+        return $this->openApi;
+    }
+
+    /**
+     * @throws ValidationFailed
+     */
+    public function validate(CallbackAddress $opAddr, RequestInterface $serverRequest) : void
+    {
+        $this->validator->validate($opAddr, $serverRequest);
+    }
+}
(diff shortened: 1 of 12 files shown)


src/PSR7/CallbackRequestValidator.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenAPIValidation\PSR7;
+
+use cebe\openapi\spec\OpenApi;
+use OpenAPIValidation\PSR7\Exception\ValidationFailed;
+use OpenAPIValidation\PSR7\Validators\BodyValidator\BodyValidator;
+use OpenAPIValidation\PSR7\Validators\CookiesValidator;
```
scaytrase: This class does not exist anymore

src/PSR7/SpecFinder.php
```diff
@@ -284,4 +300,23 @@ public function findCookieSpecs(OperationAddress $addr) : array
...
     }
+
+    /**
+     * @throws NoCallback
+     */
+    private function findCallbackInOperation(CallbackAddress $addr, Operation $operation) : Operation
+    {
+        $callbacks = $operation->callbacks;
+        if (! isset($callbacks[$addr->callbackName()])) {
+            throw NoCallback::fromCallbackPath($addr->path(), $addr->method(), $addr->callbackName(), $addr->callbackMethod());
```
scaytrase: I hope we can pass the whole `addr` here instead of destructing it to args?

src/PSR7/SpecFinder.php
```diff
@@ -215,7 +231,7 @@ public function findResponseSpec(ResponseAddress $addr) : ResponseSpec
     public function findHeaderSpecs(OperationAddress $addr) : array
     {
         // Response headers are specified differently from request headers
-        if ($addr instanceof ResponseAddress) {
+        if ($addr instanceof ResponseAddress || $addr instanceof CallbackResponseAddress) {
```
scaytrase: Not sure if it worth it, but we can try to itroduce an interface or something like this, instead of doing additional manual `assert`s  and `instanceof` checks and dropping the signatures

src/PSR7/SpecFinder.php
```diff
@@ -184,10 +192,18 @@ public function findBodySpec(OperationAddress $addr) : array
     /**
      * Find the schema which describes a given response
      *
+     * @param ResponseAddress|CallbackResponseAddress $addr
+     *
      * @throws NoPath
+     * @throws NoResponseCode
```
scaytrase: `NoPath` already covers `NoResponseCode` subclass

src/PSR7/SpecFinder.php
digilist: Yes, we could do that. It was so for all other exceptions, so I thought that there should be no dependency to the address.

src/PSR7/SpecFinder.php
digilist: I thought about the same, but then decided against it, as there will not be that many other addresses (or right now there isn't any?). Furthermore, as you said in the summary comment, this would be a bigger refactoring, so a further reason I didn't do that yet.

src/PSR7/SpecFinder.php
scaytrase: Can you please fix at least this one and rebase PR (we have updated root namespace for the package)

digilist: Sure, did it just now.

tests/PSR7/CallbackRequestTest.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenAPIValidationTests\PSR7;
```
scaytrase: Here should be new namespace too

digilist: Right, missed that one, sorry :relaxed:

digilist:
Thank you for your review and merging it! :blush:

scaytrase:
Thank you @digilist https://github.com/thephpleague/openapi-psr7-validator/releases/tag/0.5

PR #8, opened 2019-09-20
Add media-range matching support

sunspikes:
Support media ranges in response like `image/*`, `*/*`

https://tools.ietf.org/html/rfc7231#section-5.3.2

### DIFF
--- a/src/PSR7/Validators/BodyValidator/BodyValidator.php
+++ b/src/PSR7/Validators/BodyValidator/BodyValidator.php
@@ -12,6 +12,7 @@ use OpenAPIValidation\PSR7\OperationAddress;
 use OpenAPIValidation\PSR7\SpecFinder;
 use OpenAPIValidation\PSR7\Validators\ValidationStrategy;
 use Psr\Http\Message\MessageInterface;
+use function explode;
 use function preg_match;
 use function strtok;
 
@@ -49,7 +50,8 @@ final class BodyValidator implements MessageValidator
         }
 
         // does the response contain one of described media types?
-        if (! ($mediaTypeSpec = $this->matchMediaTypeSpec($mediaTypeSpecs, $contentType))) {
+        $mediaTypeSpec = $this->matchMediaTypeSpec($mediaTypeSpecs, $contentType);
+        if ($mediaTypeSpec === null) {
             throw InvalidHeaders::becauseContentTypeIsNotExpected($contentType, $addr);
         }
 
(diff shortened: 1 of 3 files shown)


src/PSR7/Validators/BodyValidator/BodyValidator.php
```diff
@@ -47,12 +49,12 @@ public function validate(OperationAddress $addr, MessageInterface $message) : vo
         }
 
         // does the response contain one of described media types?
-        if (! isset($mediaTypeSpecs[$contentType])) {
+        if (! ($mediaTypeSpec = $this->matchMediaTypeSpec($mediaTypeSpecs, $contentType))) {
```
lezhnev74: a minor opinion here - let's not assign variables in if statements?

scaytrase: Can we just tune CS checks configs for things like this?

lezhnev74: I'll review how to configure `phpcs`. Maybe it is time to introduce static analyzer which can add extra checks and QA.

tests/stubs/multi-media-types.yaml
```diff
@@ -0,0 +1,26 @@
...
+      operationId: get-multi-media-type
+      responses:
+        200:
+          description: media image
+          content:
+            'image/png':
+              schema:
+                type: string
+                format: binary
+            'image/*':
```
scaytrase: I fully understand that this approach is way better than we currently have, and following question shouldn't be the show stopper for PR but I will ask it anyway.

What if we have different schema for `type` and `type-range` like 

```
          content:
            'my-types/type1':
              schema:
                type: integer
            'my-types/*':
              schema:
                type: string
```

and we have a requst with `content-type: my-types/type1` and `string` body. Should we match second content-type instead of first here? In other words should these all behave like a combo `AnyOf` validator?

sunspikes: @scaytrase thank you for checking this PR.

Regarding your question. I think it makes sense to do the lookup like how the Accept header specification (https://tools.ietf.org/html/rfc7231#section-5.3.2)

> If more than one media range applies to a
> given type, the most specific reference has precedence

For example if your spec is:

```
content:
            'my-types/type1':
              schema:
                type: integer
            'my-types/*':
              schema:
                type: string
            '*/*':
              schema:
                type: string
```

If your response content-type is `my-types/type1` the most specific reference is 1
If your response content-type is `my-types/type2` the most specific reference is 2
If your response content-type is `different-type/sub-type` the most specific reference is 3

Hope it makes sense for you

scaytrase: I did OAS3 docs digging and it seems not really legit to have different schemas for subtypes.

https://swagger.io/docs/specification/media-types/

```
 For example, the image/* placeholder for a response body means that the server will use the same data structure for all the responses that match the placeholder
```

so maybe it's okay to process this in any way we want

src/PSR7/Validators/BodyValidator/BodyValidator.php
lezhnev74: The doctrine coding standard actually includes this check. Ref: https://github.com/doctrine/coding-standard/blob/master/lib/Doctrine/ruleset.xml see `Forbid assignments in conditions`

tests/stubs/multi-media-types.yaml
sunspikes: Yes exactly, like they mentioned in the swagger document.

In my application I have a file download endpoint and the response will be a binary file of any content type (image, video, text ...etc) and I have the following schema.

```
  '/file/{fileId}':
    get:
      tags:
        - file
      operationId: fileGet
      parameters:
        -
          name: fileId
          in: path
          required: true
          schema:
            type: integer
            format: int64
      responses:
        '200':
          description: 'File content'
          headers:
            Content-Type:
              schema:
                type: string
          content:
            '*/*':
              schema:
                type: string
                format: binary
```

Here for example i cannot specify the exact content type and this PR will make the validator validate the response properly in this scenario.

scaytrase:
Thank you @sunspikes

sunspikes:
@lezhnev74 @scaytrase thank you! It would be awesome if we could get a new release.

scaytrase:
@sunspikes https://github.com/thephpleague/openapi-psr7-validator/releases/tag/0.3 here it is

sunspikes:
thank you so much!!

PR #9, opened 2019-09-23
Fix BodyValidator for media-range scenario

sunspikes:
This fixes a bug in my previous PR https://github.com/thephpleague/openapi-psr7-validator/pull/8

This PR:

- Updates the `$mediaTypeSpec` references in `BodyValidator` in the places where it was missing
- Updates the test with a better use case (Also I just found the validation where skipped in the previous test) 

I am sorry for this, I should have doubled checked the tests. :(

### DIFF
--- a/src/PSR7/Validators/BodyValidator/BodyValidator.php
+++ b/src/PSR7/Validators/BodyValidator/BodyValidator.php
@@ -65,11 +65,11 @@ final class BodyValidator implements MessageValidator
 
         // Validate message body
         if (preg_match('#^multipart/.*#', $contentType)) {
-            (new MultipartValidator($mediaTypeSpecs[$contentType], $contentType))->validate($addr, $message);
+            (new MultipartValidator($mediaTypeSpec, $contentType))->validate($addr, $message);
         } elseif (preg_match('#^application/x-www-form-urlencoded$#', $contentType)) {
-            (new FormUrlencodedValidator($mediaTypeSpecs[$contentType], $contentType))->validate($addr, $message);
+            (new FormUrlencodedValidator($mediaTypeSpec, $contentType))->validate($addr, $message);
         } else {
-            (new UnipartValidator($mediaTypeSpecs[$contentType], $contentType))->validate($addr, $message);
+            (new UnipartValidator($mediaTypeSpec, $contentType))->validate($addr, $message);
         }
     }
 
--- a/tests/PSR7/Validators/BodyValidatorTest.php
+++ b/tests/PSR7/Validators/BodyValidatorTest.php
@@ -29,7 +29,6 @@ Content-Type: application/x-www-form-urlencoded; charset=utf-8
 address=Moscow%2C+ulitsa+Rusakova%2C+d.15&id=59731930-a95a-11e9-a2a3-2a2ae2dbcce4&phones%5B0%5D=123-456&phones%5B1%5D=456-789&phones%5B%5D=101-112
 HTTP
 ,
-                new OperationAddress('/urlencoded/scalar-types', 'post'),
             ],
             [
                 __DIR__ . '/../../stubs/multi-media-types.yaml',
@@ -53,7 +52,6 @@ Content-Length: 13
 <html></html>
 HTTP
 ,
-                new OperationAddress('/post-media-range', 'post'),
             ],
             [
                 __DIR__ . '/../../stubs/multi-media-types.yaml',
@@ -65,7 +63,6 @@ Content-Length: 1
 1
 HTTP
 ,
-                new OperationAddress('/post-media-range', 'post'),
             ],
         ];
     }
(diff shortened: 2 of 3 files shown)


tests/PSR7/Validators/BodyValidatorTest.php
```diff
@@ -76,7 +84,7 @@ public function testValidateGreen(string $specFile, string $message) : void
         );
 
         $validator = (new ValidatorBuilder())->fromYamlFile($specFile)->getServerRequestValidator();
-        $validator->validate($serverRequest);
-        $this->addToAssertionCount(1);
+        $opAddress = $validator->validate($serverRequest);
+        $this->assertEquals($expectedOpAddress, $opAddress);
```
scaytrase: Do we really need this assertion? In case we have only one op addr in fixture schema not catching an exception here means that we matched that single one

src/PSR7/Validators/BodyValidator/BodyValidator.php
```diff
@@ -65,11 +65,11 @@ public function validate(OperationAddress $addr, MessageInterface $message) : vo
 
         // Validate message body
         if (preg_match('#^multipart/.*#', $contentType)) {
-            (new MultipartValidator($mediaTypeSpecs[$contentType], $contentType))->validate($addr, $message);
+            (new MultipartValidator($mediaTypeSpec, $contentType))->validate($addr, $message);
         } elseif (preg_match('#^application/x-www-form-urlencoded$#', $contentType)) {
-            (new FormUrlencodedValidator($mediaTypeSpecs[$contentType], $contentType))->validate($addr, $message);
+            (new FormUrlencodedValidator($mediaTypeSpec, $contentType))->validate($addr, $message);
```
scaytrase: In this place this doesn't play any difference, that's why tests passed this. Since `preg_match` checks the strict equality here.

src/PSR7/Validators/BodyValidator/BodyValidator.php
```diff
@@ -65,11 +65,11 @@ public function validate(OperationAddress $addr, MessageInterface $message) : vo
 
         // Validate message body
         if (preg_match('#^multipart/.*#', $contentType)) {
-            (new MultipartValidator($mediaTypeSpecs[$contentType], $contentType))->validate($addr, $message);
+            (new MultipartValidator($mediaTypeSpec, $contentType))->validate($addr, $message);
```
scaytrase: Can we add more tests on this line to be sure they are behave correctly using range?

scaytrase:
> This fixes a bug in my previous PR #8

as far as I understand this only belongs to new media-range feature and all old code works correctly, yes while there is strictly matching content-type in schema?

src/PSR7/Validators/BodyValidator/BodyValidator.php
sunspikes: The earlier tests passed because, 

This is a server request validator, the openapi schema in the tests had media-range specified for response schema and the the response schema was never validated.

Now I moved the media-ranges to `requestBody` so that it will be validated by the server request validator

src/PSR7/Validators/BodyValidator/BodyValidator.php
sunspikes: Do we need this because the issue was not because of the strict comparison as i mentioned above 🤔

sunspikes:
> > This fixes a bug in my previous PR #8
> 
> as far as I understand this only belongs to new media-range feature and all old code works correctly, yes? while there is strictly matching content-type in schema

Yes exactly because `$mediaTypeSpecs[$contentType]` will still get the schema and `$mediaTypeSpec` needed only for media-range case

src/PSR7/Validators/BodyValidator/BodyValidator.php
scaytrase: Ah, got it

sunspikes:
> We need to involve coverage checks in order not to miss things like this, I think

Yes that would be awesome. code Climate is pretty good and easy to configure https://docs.codeclimate.com/docs/configuring-test-coverage

sunspikes:
@scaytrase thank you so much for the quick turnaround! can you please also add a release with this fix?

scaytrase:
already done as 0.3.1

scaytrase:
> Yes that would be awesome. code Climate is pretty good and easy to configure https://docs.codeclimate.com/docs/configuring-test-coverage

@lezhnev74 do you have permissions to configure things like this here? Because I don't

lezhnev74:
nope, but @frankdejonge should have it

frankdejonge:
Would scrutinizer-ci work as as well?

scaytrase:
I hope scrutinizer is good too, using it for own projects

frankdejonge:
So far so good: https://scrutinizer-ci.com/g/thephpleague/openapi-psr7-validator/inspections/c9b71a63-4de3-4729-8038-a7c7ceadf1be

scaytrase:
Nice, I'll tune repo-level config to enable coverage a bit later. Thanks, @frankdejonge !

frankdejonge:
You're welcome! 👍keep up the good work!

PR #10, opened 2019-09-23
Fix uniqueitems validator for multi-level items

sunspikes:
An array can contain list of `integer`, `number`, `boolean`, `string`, `object` or other `array`s

https://swagger.io/docs/specification/data-models/data-types/#uniqueItems

Currenty the `uniqueItems` validator works only for array of `integer`, `number`, `boolean` or `string` and not for array of `object`s or other `array`s

This PR fixes the validator to handles those scenarios as well. Also i have added tests for more scenarios.

### DIFF
--- a/tests/Schema/Keywords/UniqueItemsTest.php
+++ b/tests/Schema/Keywords/UniqueItemsTest.php
@@ -52,6 +52,17 @@ SPEC
 schema:
   type: array
   items:
+    type: number
+  uniqueItems: true
+SPEC
+,
+                [1, 1.0],
+            ],
+            [
+                <<<SPEC
+schema:
+  type: array
+  items:
     type: boolean
   uniqueItems: true
 SPEC


src/Schema/Keywords/UniqueItems.php
```diff
@@ -39,7 +42,14 @@ public function validate($data, bool $uniqueItems) : void
             throw InvalidSchema::becauseDefensiveSchemaValidationFailed($e);
         }
 
-        if (count(array_unique($data)) !== count($data)) {
+        $items = $data;
+        if (is_array($data) && count($data)) {
```
scaytrase: I don't really thing that we need both checks here. It's known it is array data, because of assertion above. And if we won't check count than we will just get empty array if it's initially empty. So we can omit this check entirely and get more simple code

src/Schema/Keywords/UniqueItems.php
```diff
@@ -39,7 +42,14 @@ public function validate($data, bool $uniqueItems) : void
...
+            $items = array_map(static function ($item) {
+                return var_export($item, true);
```
scaytrase: This approach has a flaw in handling numerics. Imagine we have an `array of floats` spec. Eval the followin snippet to ensure your code consider `[1, 1.0]` as to unique items, while current should tell they are the same. You can add a test with this case too

```
function export($val)
{
    echo var_export($val) . PHP_EOL;
}

$data = [1, 1.0];
// this output two different strings which would be considered different later
array_map('export', $data);
// this says there is only one unique element in array
var_dump(array_unique($data));
```

sunspikes: Thank you for noticing this. I just had a look into the other json schema v4 validators, like

https://github.com/justinrainbow/json-schema
https://www.npmjs.com/package/jsonschema

I see those are handling integers and floats differently in a `number` array. that is if you have a number array `[1, 1.0]` it will be unique.

Screen Shot 2019-09-24 at 11 31 17

Not sure if this behavior is fine or should we handle it differently, like it works with python json schema validator

https://pypi.org/project/jsonschema/

sunspikes: Thank you for noticing this,

So are you saying it should be working as it is with the current code. That is `[1, 1.0]` is not unique?

Because this works differently in different validators, for example,

https://www.npmjs.com/package/jsonschema and https://github.com/justinrainbow/json-schema

validates the above array to be unique (that is, 1 !== 1.0),

where as validators in other languages which has native float type like python and ruby, for example https://pypi.org/project/jsonschema/ validates the above not unique.

scaytrase: I think as an end-user I expect that if I had `uniqueItems: true` and data passed validation i would expect things like inserting this values into DB with unique constraint passing. As soon as we make a validation library I think we should rely on how we expect users utilize the validated data. But it also good thing to look how other deal such problems, for sure, as soon it could be more expected behaviour

@lezhnev74 WDYT?

lezhnev74:
I've merged PR which fixes some invalid type checks.

This PR makes sense as it is now:

This schema would fail for `[1, 1.0]` as these are not integers
```
[
                <<<SPEC
schema:
  type: array
  items:
    type: integer
  uniqueItems: true
SPEC
,
                [1, 1.0],
            ],
```



This schema would succeed for `[1, 1.0]` as these are unique numbers of different types
```
[
                <<<SPEC
schema:
  type: array
  items:
    type: number
  uniqueItems: true
SPEC
,
                [1, 1.0],
            ],
```

sunspikes:
@scaytrase can you also please create release

scaytrase:
Sure, https://github.com/thephpleague/openapi-psr7-validator/releases/tag/0.3.2

PR #12, opened 2019-09-24
Add missing failing tests

lezhnev74:
While reading #10 I noticed some edge cases are not properly validated in `\OpenAPIValidation\Schema\Keywords\Type`.

Added two tests which are failing now (we missed them):
- 1.0 should not be validated as integer
- [1,2] should not be validated as boolean


lezhnev74:
Extra tests for validation booleans and integers are green now

PR #15, opened 2019-10-01
Use `crc32b` instead of crc32 function

sunspikes:
Because `crc32()` is platform dependant and is not portable between 32 and 64 bit platforms, it produces a signed int, see: https://www.php.net/manual/en/function.crc32.php

`hash('crc32b', $content)` is consistent across platforms (Also I see it produces same as in other languages like golang and java)


lezhnev74:
Nice catch!

sunspikes:
🎉

PR #21, opened 2019-10-11
Allow milliseconds in date-time format. Fixes #19

### LINKED ISSUE #19: Date-time format should allow optional milliseconds

coatesap:
Fixes #19 

Although #20 has also been submitted to fix the same issue, as I'd already written this I'll push it up and let the maintainers decide which one to use.

### DIFF
--- a/src/Schema/TypeFormats/StringDateTime.php
+++ b/src/Schema/TypeFormats/StringDateTime.php
@@ -10,8 +10,9 @@ class StringDateTime
 {
     public function __invoke(string $value) : bool
     {
-        // the date-time notation as defined by RFC 3339, section 5.6, for example, 2017-07-21T17:32:28Z
-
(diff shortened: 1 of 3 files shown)


coatesap:
Retracting PR, as I didn't expect anyone else to take an interest in this issue (!), but #20 seems a more comprehensive fix.

Issue #30, opened 2019-10-29
Move code to the namespace under `League\`


lezhnev74:
The league packages tend to use this root namespace, so we should too. Should be pretty straightforward thing with help of IDE.

If no one volunteers to do that, I will this weekend.

scaytrase:
Yea, since we are still under 0.x version we can afford this

PR #31, opened 2019-10-29
Move code to the namespace under `League\`

sunspikes:
Fixes https://github.com/thephpleague/openapi-psr7-validator/issues/30

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,12 +12,12 @@
     "type": "library",
     "autoload": {
         "psr-4": {
-            "OpenAPIValidation\\": "src/"
+            "League\\OpenAPIValidation\\": "src/"
         }
     },
     "autoload-dev": {
         "psr-4": {
-            "OpenAPIValidationTests\\": "tests/"
+            "League\\OpenAPIValidationTests\\": "tests/"
         }
     },
     "require": {
--- a/src/Foundation/ArrayHelper.php
+++ b/src/Foundation/ArrayHelper.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace OpenAPIValidation\Foundation;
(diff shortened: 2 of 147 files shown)


lezhnev74:
Well, checks are green. @scaytrase go ahead with reviewing / merging / tagging if all good.

scaytrase:
Sure, will do as soon as I can

composer.json
```diff
@@ -12,12 +12,12 @@
...
     "autoload": {
         "psr-4": {
-            "OpenAPIValidation\\": "src/"
+            "League\\OpenAPIValidation\\": "src/"
         }
     },
     "autoload-dev": {
         "psr-4": {
-            "OpenAPIValidationTests\\": "tests/"
+            "League\\OpenAPIValidationTests\\": "tests/"
```
scaytrase: @lezhnev74 how about `League\OpenAPIValidation\Tests\` for test to have consistent `League\OpenAPIValidation` as base namespace?

sunspikes: @scaytrase Imho it should be fine either way. Afaics, all league libraries do it differently, for example:

https://github.com/thephpleague/flysystem/blob/master/tests/AdaptersThatCanOverwriteFilesTest.php
https://github.com/thephpleague/tactician/blob/master/tests/CommandBusTest.php
https://github.com/thephpleague/oauth2-server/blob/master/tests/AuthorizationServerTest.php

lezhnev74: Yeah, good point. Checked another package, it does the same naming:
https://github.com/thephpleague/commonmark/blob/master/composer.json

sunspikes: @lezhnev74 / @scaytrase  would you prefer changing it that way?

PR #33, opened 2019-11-06
Fix validation of empty objects

digilist:
As the JSON payload is decoded into arrays, an empty array is not considered valid
for an object, even though it might come from an object.

https://github.com/thephpleague/openapi-psr7-validator/blob/master/src/PSR7/Validators/BodyValidator/UnipartValidator.php#L45

I think this broke after #24

### DIFF
--- a/src/Schema/Keywords/Type.php
+++ b/src/Schema/Keywords/Type.php
@@ -42,7 +42,7 @@ class Type extends BaseKeyword
     {
         switch ($type) {
             case CebeType::OBJECT:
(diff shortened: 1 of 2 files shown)


scaytrase:
Good catch, thank you! we should move to using `\stdClass` everywhere to represent objects for sure to avoid such errors

scaytrase:
thank you @digilist https://github.com/thephpleague/openapi-psr7-validator/releases/tag/0.5.1

PR #36, opened 2019-11-27
Rename variable in "Response message" chapter

markuspoerschke:
Since the HTTP response should be validated, the variable should be called `$response` instead of `$request`.


scaytrase:
Thanks @markuspoerschke

PR #40, opened 2019-12-09
Path finder in spec finder

### LINKED ISSUE #42: Spec Finder Not Always Finding
I found that SpecFinder as it was trying to find paths based on pure string matching which doesnt work if the path contains paramters

this PR fixes that: https://github.com/thephpleague/openapi-psr7-validator/pull/40

edmondscommerce:
For paths with parameters, spec finder was failing - this fixes that

fixes #42

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,45 +1,46 @@
 {
-    "name": "league/openapi-psr7-validator",
-    "description": "Validate PSR-7 messages against OpenAPI (3.0.2) specifications \nexpressed in YAML or JSON",
-    "keywords": [
-        "OpenAPI",
-        "psr7",
-        "http",
-        "validation"
-    ],
-    "homepage": "https://github.com/thephpleague/openapi-psr7-validator",
-    "license": "MIT",
-    "type": "library",
-    "autoload": {
-        "psr-4": {
-            "League\\OpenAPIValidation\\": "src/"
-        }
-    },
-    "autoload-dev": {
-        "psr-4": {
-            "League\\OpenAPIValidation\\Tests\\": "tests/"
-        }
-    },
-    "require": {
-        "php": ">=7.2",
-        "ext-json": "*",
-        "cebe/php-openapi": "^1.3",
-        "dflydev/fig-cookies": "^2.0",
-        "psr/cache": "^1.0",
-        "psr/http-message": "^1.0",
-        "psr/http-server-middleware": "^1.0",
-        "respect/validation": "^1.1",
-        "riverline/multipart-parser": "^2.0.3",
-        "webmozart/assert": "^1.4"
-    },
-    "require-dev": {
-        "cache/array-adapter": "^1.0",
-        "doctrine/coding-standard": "^6.0",
-        "guzzlehttp/psr7": "^1.5",
-        "phpunit/phpunit": "^7|^8"
-    },
-    "config": {
-        "sort-packages": true
-    },
-    "prefer-stable": true
+  "name": "league/openapi-psr7-validator",
+  "description": "Validate PSR-7 messages against OpenAPI (3.0.2) specifications \nexpressed in YAML or JSON",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -1,45 +1,47 @@
...
-        "doctrine/coding-standard": "^6.0",
-        "guzzlehttp/psr7": "^1.5",
-        "phpunit/phpunit": "^7|^8"
-    },
-    "config": {
-        "sort-packages": true
-    },
-    "prefer-stable": true
+  "name": "league/openapi-psr7-validator",
+  "description": "Validate PSR-7 messages against OpenAPI (3.0.2) specifications \nexpressed in YAML or JSON",
```
thunderer: @edmondscommerce any reason for reformatting the file?

thunderer:
Hi @edmondscommerce, thanks for this PR, it would be really nice if it could make it into library soon. I assume that most of our endpoints contain parameters, so that would make this library way more useful. :) Keep up the good work!

composer.json
edmondscommerce: muscle memory I'm afraid - I always reformat before save

is it a big deal?

thunderer: I don't think so, the feature in this PR is way more important. :) Can we merge it?

scaytrase: The problem is that since there is no `.editorconfig` in project - next contributor could reformat it back. It's not really a big problem but I'd rather suggest to keep meaningful changes only

src/PSR7/SpecFinder.php
```diff
@@ -74,37 +108,49 @@ public function findPathSpec(OperationAddress $addr) : PathItem
         return $pathSpec;
     }
 
-    /**
-     * @return Parameter[]
-     *
-     * @throws NoPath
-     */
-    public function findPathSpecs(OperationAddress $addr) : array
+    private function findPathSimple(OperationAddress $addr) : ?PathItem
```
scaytrase: I'd rather suggest to name it something like `findExactPath` and `findMatchingTemplatePath` or something like this. Current naming gives no real understanding what each method does under the hood.

composer.json
```diff
@@ -1,45 +1,47 @@
...
+      "League\\OpenAPIValidation\\Tests\\": "tests/"
+    }
+  },
+  "require": {
+    "php": ">=7.2",
+    "ext-json": "*",
+    "cebe/php-openapi": "^1.3",
+    "dflydev/fig-cookies": "^2.0",
+    "psr/cache": "^1.0",
+    "psr/http-factory": "^1.0",
```
scaytrase: can't find any usages

src/PSR7/SpecFinder.php
scaytrase: Also we can try incapsulate this method into PathFinder, keeping thing simpler.

src/PSR7/SpecFinder.php
```diff
@@ -74,37 +108,49 @@ public function findPathSpec(OperationAddress $addr) : PathItem
...
+        $results = $finder->search();
+        if (count($results) === 1) {
+            return $this->findPathSimple($results[0]);
+        }
 
-        foreach ($spec->parameters as $p) {
-            if ($p->in !== 'path') {
-                continue;
-            }
+        return null;
```
scaytrase: It could turn out that there are multiple matched paths (more than one). returning null here means throwing `NoPath` exception later. Sound non-intuitive. I think we should check this case too

scaytrase:
@edmondscommerce do you have plans on continuing this PR?

edmondscommerce:
Hi

I'm afraid the project this was for has now gone live with my fork for now.

I'll close the PR on the basis that I'm more than happy for you to adjust the branch to your requirements and merge it in. I'm in the post Christmas overload at the moment so I don't see me having the free time to look into this in the near future

PR #41, opened 2019-12-09
Handle missing params

### LINKED ISSUE #39: Path Validator Not Handling Missing Params
Firstly - thanks for this great library!

Found a fairly easy to fix issue though I think:

It's quite easy to get an undefined index error at 

https://github.com/thephpleague/openapi-psr7-validator/blob/master/src/PSR7/Validators/PathValidator.php#L53

if you have not done written your spec properly, ie you have a path with a `{param}` but haven't actually added the param annotation to the request spec

This seems like an easy issue to catch and throw a useful exception as is being done elsewhere

I'd be happy to give it a go as a PR

edmondscommerce:
Fixes #39

### DIFF
--- a/src/PSR7/Validators/PathValidator.php
+++ b/src/PSR7/Validators/PathValidator.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace League\OpenAPIValidation\PSR7\Validators;
 
+use League\OpenAPIValidation\PSR7\Exception\NoParameter;
 use League\OpenAPIValidation\PSR7\Exception\NoPath;
 use League\OpenAPIValidation\PSR7\Exception\Validation\InvalidPath;
 use League\OpenAPIValidation\PSR7\MessageValidator;
@@ -27,9 +28,9 @@ final class PathValidator implements MessageValidator
     }
 
     /** {@inheritdoc} */
-    public function validate(OperationAddress $addr, MessageInterface $message) : void
+    public function validate(OperationAddress $addr, MessageInterface $message): void
     {
-        if (! ($message instanceof RequestInterface)) {
+        if (!($message instanceof RequestInterface)) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


src/PSR7/Validators/PathValidator.php
```diff
@@ -39,14 +40,21 @@ public function validate(OperationAddress $addr, MessageInterface $message) : vo
...
         $specs = $this->finder->findPathSpecs($addr);
 
         $path             = $message->getUri()->getPath();
         $pathParsedParams = $addr->parseParams($path); // ['id'=>12]
-        $validator        = new SchemaValidator($this->detectValidationStrategy($message));
+        if ($pathParsedParams===[]) {
+            return;
+        }
+        if ($specs === []) {
+            throw NoParameter::fromPath($addr->path());
```
scaytrase: This tends to be not a `ValidationFailed` exception, but rather and `InvalidSchema` since every parameter in path should be described. But speaking more general passing this point means that we are validating address which doesn't match any operation in our spec and there is nothing to do with missing params.

tests/PSR7/Validators/PathValidatorTest.php
```diff
@@ -0,0 +1,39 @@
...
+use GuzzleHttp\Psr7\Uri;
+use League\OpenAPIValidation\PSR7\Exception\ValidationFailed;
+use League\OpenAPIValidation\PSR7\ValidatorBuilder;
+use PHPUnit\Framework\TestCase;
+
+class PathValidatorTest extends TestCase
+{
+    public function testItThrowsHelpfulExceptionWhenMissingParams() : void
+    {
+        $this->expectException(ValidationFailed::class);
```
scaytrase: In this given example there is no `ValidationFailed` expectation. The given schema is invalid since it's not described in spec. Path parameters at least should have `required: true` as there are in path

src/PSR7/Validators/PathValidator.php
```diff
@@ -39,14 +40,21 @@ public function validate(OperationAddress $addr, MessageInterface $message) : vo
...
+        }
+        $validator = new SchemaValidator($this->detectValidationStrategy($message));
 
         foreach ($pathParsedParams as $name => $value) {
```
scaytrase: It looks like that this PR doesn't address the problem described in the original issue. If we have two path parameters, and one of them is described and a second one is not - we will still reach this point having `$specs[$name]` index problems.

Having no specs from non-empty params is bad for sure but the problem that needs to be fixed is that we are missing specific paramter in schema, an this is not checked.

scaytrase:
@edmondscommerce do you have plans on continuing this PR?

edmondscommerce:
Hi 

I'm afraid the project this was for has now gone live with my fork for now. 

I'll close the PR on the basis that I'm more than happy for you to adjust the branch to your requirements and merge it in. I'm in the post Christmas overload at the moment so I don't see me having the free time to look into this in the near future

Issue #2, opened 2019-09-19
Add parameters serialisation support


lezhnev74:
Currently, the package does not support serialization of parameters.

This ticket is to track efforts for adding serialization support to the package.

### Docs:
https://swagger.io/docs/specification/serialization/
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#support-for-x-www-form-urlencoded-request-bodies

scaytrase:
To track progress: 
@samuelnogueira implemented JSON parameter serialization in #43

lezhnev74:
I am going to add the missing support for validation of serialized parameters.
After 1st iteration here, I think [this test](https://github.com/thephpleague/openapi-psr7-validator/blob/16b0e9e649a0e9fe8d9cb6993ef9c38a9198bd44/tests/PSR7/Validators/RequestParameterTest.php#L23) to be removed once [this PR](https://github.com/cebe/php-openapi/pull/53) is merged to underlying cebe package (as we should delegate schema validation to that package and not doing it in this one, right?).

lezhnev74:
Did you know, by default PHP would override get arguments with the same name. However it is allowed in general by OpenAPI spec.

For get parameters like `?id=3&id=4&id=5`,  the value of `$_GET` would be `id=5`. Yet, `$_SERVER['REQUEST_URI']` still would contain the initial query string. 

I think, query string discovery is beyond this package's responsibility, it works with what it is given (psr-7 object).

lezhnev74:
Quick update: awesome @cebe merged [the pending PR](https://github.com/cebe/php-openapi/pull/53) and I am ready to move to the next step here.

scaytrase:
Just merged #72 with support for some serializaition styles

PR #43, opened 2019-12-19
Added support for JSON query parameters

samuelnogueira:
This is part of the effort required for #2 

Only adds support for JSON serialized query parameters, no support for other forms of serialization (ex. application/xml) were added, or other types of parameters (ex. cookies).

A use case for this type of parameters is used as example in OpenAPI documentation:
https://swagger.io/docs/specification/describing-parameters/#schema-vs-content

### DIFF
--- a/src/PSR7/Validators/QueryArgumentsValidator.php
+++ b/src/PSR7/Validators/QueryArgumentsValidator.php
@@ -16,8 +16,13 @@ use League\OpenAPIValidation\Schema\SchemaValidator;
 use Psr\Http\Message\MessageInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use const JSON_ERROR_NONE;
 use function array_key_exists;
+use function json_decode;
+use function json_last_error;
+use function json_last_error_msg;
 use function parse_str;
+use function preg_match;
 
 /**
  * @see https://swagger.io/docs/specification/describing-parameters/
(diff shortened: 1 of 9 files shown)


src/PSR7/Validators/QueryArgumentsValidator.php
```diff
@@ -75,20 +80,33 @@ private function checkMissingArguments(OperationAddress $addr, array $parsedQuer
...
+            // skip if there is no spec for this argument
             if (! array_key_exists($name, $specs)) {
                 continue;
             }
 
+            $data      = $argumentValue;
+            $parameter = $specs[$name];
+            $schema    = $parameter->schema;
+            if ($schema === null) {
+                // There should be one and only one entry in 'content'.
```
scaytrase: It sounds that we rely on some kind of docs here. Can you share a link to it? What happens if we have zero entries? two entries?

src/PSR7/Validators/QueryArgumentsValidator.php
```diff
@@ -108,4 +126,21 @@ private function parseQueryArguments(RequestInterface $message) : array
...
+     * @return mixed
+     *
+     * @throws InvalidQueryArgs
+     */
+    private static function parseArgumentValue(string $value, string $contentType, OperationAddress $addr, string $argumentName)
+    {
+        if (preg_match('#^application/.*json$#', $contentType)) {
+            $value = json_decode($value, true);
+            if (json_last_error() !== JSON_ERROR_NONE) {
+                throw InvalidQueryArgs::becauseValueIsNotValidJson(json_last_error_msg(), $argumentName, $addr);
```
scaytrase: How about more generic `becauseDoesNotMatchContentType` or something like this? We can reuse it later for XML or other types

tests/FromCommunity/JsonQueryParameterValidationTest.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\OpenAPIValidation\Tests\FromCommunity;
```
scaytrase: I hope "FromCommunity" namespace exist to pin reproducing cases from different issues discussions into test flow. Since this is a general feature implementation, I suggest you to relocate this code to namespace related to new featuers

src/PSR7/Validators/QueryArgumentsValidator.php
samuelnogueira: > It sounds that we rely on some kind of docs here. Can you share a link to it?

I found strict rules on how to use schema/content in [Swagger Editor](https://editor.swagger.io/). By the magic of testing all possible combinations, I found these are considered structural errors:
1. Having both `schema` and `content`
2. Not having `schema` or `content`
3. Having `content` with no properties
4. Having `content` with more than 1 properties

The documentation in Swagger website is a bit vague here, I could only find this relevant quote:
https://swagger.io/docs/specification/describing-parameters/#schema-vs-content
> To describe the parameter contents, you can use either the schema or content keyword. They are mutually exclusive (...)

It looks good enough to me, so I'll add a link to it in the comments., 

> What happens if we have zero entries? two entries?

Both these cases are invalid OpenAPI specifications, however, `cebe/php-openapi` doesn't raise any issues while parsing files with these problems. I also couldn't find in this library (`thephpleague/openapi-psr7-validator`) any exceptions related to errors in the OpenAPI specification itself.
To avoid running into the null type error again, I would suggest one of the following:
- Throw an Exception (ex. InvalidSpecification) that does **not** extend from `ValidationFailed` for any of the 4 invalid combinations I listed above
- Skip parameters that cannot be validated because the specification is invalid.

What do you think it's best here? (I'll implement one of these in this PR to show how it would look like).

scaytrase: Yea, I've checked editor too, we should throw invalid schema for count !== 1

src/PSR7/Validators/RequestParameter.php
```diff
@@ -0,0 +1,95 @@
...
+                throw ContentTypeMismatch::fromContentType($this->contentType, $value);
+            }
+
+            return $decodedValue;
+        }
+
+        return $value;
+    }
+
+    private function isJsonContentType() : bool
```
scaytrase: As a future improvement (not required for this PR) we can implement typed extensions for this class, like JsonRequestParameter, PlainRequestParameter instead of passing content-type and checking it on each deserialization

scaytrase:
Sorry for late review, great work @samuelnogueira !

samuelnogueira:
@scaytrase it's holiday season, the review wasn't late at all! Thank you for taking the time to review and merge the PR, it's much appreciated.

Issue #13, opened 2019-09-27
Missing inner exception message/detail when using AnyOf/OneOf


scaytrase:
See https://github.com/lezhnev74/openapi-psr7-validator/issues/61 for details and discussion

scaytrase:
Closed with #46

PR #46, opened 2020-01-21
Keep inner exceptions if oneOf or anyOf fails

ckdot:
this will solve https://github.com/thephpleague/openapi-psr7-validator/issues/13

### DIFF
--- a/src/Schema/Exception/NotEnoughValidSchemas.php
+++ b/src/Schema/Exception/NotEnoughValidSchemas.php
@@ -6,7 +6,6 @@ namespace League\OpenAPIValidation\Schema\Exception;
 
 use Throwable;
 
-// Indicates that data was not matched against a combination of schema keywords
 class NotEnoughValidSchemas extends KeywordMismatch
 {
     /** @var Throwable[] */
@@ -22,10 +21,9 @@ class NotEnoughValidSchemas extends KeywordMismatch
         string $keyword,
         $data,
         array $innerExceptions,
-        ?string $message = null,
-        ?Throwable $prev = null
+        ?string $message = null
     ) : KeywordMismatch {
-        $instance                  = new self('Keyword validation failed: ' . $message, 0, $prev);
+        $instance                  = new self('Keyword validation failed: ' . $message, 0);
         $instance->keyword         = $keyword;
         $instance->data            = $data;
         $instance->innerExceptions = $innerExceptions;
--- a/src/Schema/Exception/NotEnoughValidSchemas.php
+++ b/src/Schema/Exception/NotEnoughValidSchemas.php
@@ -7,7 +7,7 @@ namespace League\OpenAPIValidation\Schema\Exception;
 use Throwable;
 
 // Indicates that data was not matched against a combination of schema keywords
-class InvalidSchemaCombination extends KeywordMismatch
+class NotEnoughValidSchemas extends KeywordMismatch
 {
     /** @var Throwable[] */
     protected $innerExceptions = [];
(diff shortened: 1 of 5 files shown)


src/Schema/Keywords/OneOf.php
```diff
@@ -55,17 +56,24 @@ public function validate($data, array $oneOf) : void
...
             try {
                 $schemaValidator->validate($data, $schema, $this->dataBreadCrumb);
                 $matchedCount++;
             } catch (SchemaMismatch $e) {
-                // that did not match... its ok
+                $innerExceptions[] = $e;
             }
         }
 
         if ($matchedCount !== 1) {
```
scaytrase: For new exception we should distinguish between zero and >1 matches. For zero matches - use aggregated exception. for >1 we should just provide some useful info, since there is nothing bad with unmatched ones

ckdot: thanks for fast reply. what do you thin about this:
- keeping logic like it is for zero matches
- add another exception type "TooManyValidSchemas"* that works like InvalidSchemaCombination but instead of $innerExceptions it stores an array of $validSchemas

* if you have an idea for some better name please tell me.

scaytrase: `InvalidSchemaCombination` naming should be changed to `NotEnoughValidSchemas` then or something similar, since current naming covers both cases. All the other parts of the idea sounds good.

ckdot: i see, that makes sense. i've updated the code. feel free to check it again :)

src/Schema/Exception/NotEnoughValidSchemas.php
```diff
@@ -0,0 +1,43 @@
...
+     * @param Throwable[] $innerExceptions
+     *
+     * @return self
+     */
+    public static function fromKeywordWithInnerExceptions(
+        string $keyword,
+        $data,
+        array $innerExceptions,
+        ?string $message = null,
+        ?Throwable $prev = null
```
scaytrase: should we care about `prev` here? make no sense for me

src/Schema/Exception/TooManyValidSchemas.php
```diff
@@ -0,0 +1,44 @@
...
+     * @param Schema[] $validSchemas
+     *
+     * @return self
+     */
+    public static function fromKeywordWithValidSchemas(
+        string $keyword,
+        $data,
+        array $validSchemas,
+        ?string $message = null,
+        ?Throwable $prev = null
```
scaytrase: Same question here for prev. It should no prev exceptions by design, I hope

src/Schema/Exception/TooManyValidSchemas.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\OpenAPIValidation\Schema\Exception;
+
+use cebe\openapi\spec\Schema;
+use Throwable;
+
+// Indicates that data was not matched against a combination of schema keywords
```
scaytrase: I think comment is outdated. Thats the problem of all these comments. maybe drop them?

src/Schema/Exception/NotEnoughValidSchemas.php
ckdot: you're right; removed it.

src/Schema/Exception/TooManyValidSchemas.php
ckdot: you're right; removed it.

src/Schema/Exception/TooManyValidSchemas.php
ckdot: you're right; removed it.

scaytrase:
Thanks @ckdot !

PR #47, opened 2020-02-04
Remove note on lack of `RequestInterface` support

dabernathy89:
It looks like support for `\Psr\Http\Message\RequestInterface` was added, but the old notice about lack of support is still there.


scaytrase:
Good catch, thank you @dabernathy89 !

PR #49, opened 2020-02-20
Strict types validation (fix #48)

### LINKED ISSUE #48: String with stringified value passes validation as type: number and type: boolean
https://swagger.io/docs/specification/data-models/data-types/#numbers clearly states that string containing number is considered as string, not as number.

Same for boolean

m0003r:
Resolves #48 

Validation is now strict, but query, path and cookie params (that are always strings) attempted to convert to specified type.

### DIFF
--- a/src/PSR7/Validators/ArrayValidator.php
+++ b/src/PSR7/Validators/ArrayValidator.php
@@ -7,7 +7,6 @@ namespace League\OpenAPIValidation\PSR7\Validators;
 use cebe\openapi\spec\Parameter;
 use League\OpenAPIValidation\PSR7\Exception\Validation\InvalidParameter;
 use League\OpenAPIValidation\PSR7\Exception\Validation\RequiredParameterMissing;
-use League\OpenAPIValidation\PSR7\OperationAddress;
 use League\OpenAPIValidation\Schema\BreadCrumb;
 use League\OpenAPIValidation\Schema\Exception\SchemaMismatch;
 use League\OpenAPIValidation\Schema\SchemaValidator;
@@ -35,11 +34,11 @@ class ArrayValidator
      * @throws InvalidParameter
      * @throws RequiredParameterMissing
      */
-    public function validateArray(OperationAddress $addr, array $params, int $validationStrategy) : void
+    public function validateArray(array $params, int $validationStrategy) : void
     {
         foreach ($this->specs as $name => $spec) {
             if ($spec->required && ! array_key_exists($name, $params)) {
-                throw RequiredParameterMissing::fromNameAndAddr($name, $addr);
+                throw RequiredParameterMissing::fromName($name);
             }
         }
 
@@ -56,7 +55,7 @@ class ArrayValidator
             try {
                 $validator->validate($parameter->deserialize($argumentValue), $parameter->getSchema(), new BreadCrumb($name));
             } catch (SchemaMismatch $e) {
-                throw InvalidParameter::becauseValueDidNotMatchSchema($name, $argumentValue, $addr, $e);
+                throw InvalidParameter::becauseValueDidNotMatchSchema($name, $argumentValue, $e);
             }
         }
     }
--- a/src/PSR7/Validators/ArrayValidator.php
+++ b/src/PSR7/Validators/ArrayValidator.php
@@ -1,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\OpenAPIValidation\PSR7\Validators;
+
+use cebe\openapi\spec\Parameter;
+use League\OpenAPIValidation\PSR7\Exception\Validation\InvalidParameter;
+use League\OpenAPIValidation\PSR7\Exception\Validation\RequiredParameterMissing;
+use League\OpenAPIValidation\PSR7\OperationAddress;
+use League\OpenAPIValidation\Schema\BreadCrumb;
+use League\OpenAPIValidation\Schema\Exception\SchemaMismatch;
+use League\OpenAPIValidation\Schema\SchemaValidator;
+use function array_key_exists;
+
+/**
+ * Validates given array against given specs
+ */
+class ArrayValidator
+{
+    /** @var Parameter[] */
+    private $specs;
+
+    /**
+     * @param Parameter[] $specs
+     */
+    public function __construct(array $specs)
+    {
+        $this->specs = $specs;
+    }
+
+    /**
+     * @param mixed[] $params
+     *
+     * @throws InvalidParameter
+     * @throws RequiredParameterMissing
+     */
+    public function validateArray(OperationAddress $addr, array $params, int $validationStrategy) : void
+    {
+        foreach ($this->specs as $name => $spec) {
+            if ($spec->required && ! array_key_exists($name, $params)) {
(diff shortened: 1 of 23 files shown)


scaytrase:
Althougth the proposal itself looks valid I suggest to add more tests on HTTP related stuff. Because the problem I see the that, for example, path and query parameters are always strings by the nature of HTTP and how PHP process it. So considering any "number string" or "boolean string" to be string will break these things

src/Schema/Keywords/Type.php
```diff
@@ -52,19 +49,17 @@ public function validate($data, string $type, ?string $format = null) : void
                 }
                 break;
             case CebeType::BOOLEAN:
-                $stringifiedBool = is_scalar($data) && preg_match('#^(true|false)$#i', (string) $data);
-                if (! is_bool($data) && ! $stringifiedBool) {
+                if (is_string($data) || ! is_bool($data)) {
```
scaytrase: Looks redundant. If data is string it should be !is_bool as well, so right part is always true when left part is true and we can keep just right part

src/Schema/Keywords/Type.php
```diff
@@ -52,19 +49,17 @@ public function validate($data, string $type, ?string $format = null) : void
...
             case CebeType::NUMBER:
-                if (! is_numeric($data)) {
+                if (is_string($data) || ! is_numeric($data)) {
                     throw TypeMismatch::becauseTypeDoesNotMatch(CebeType::NUMBER, $data);
                 }
                 break;
             case CebeType::INTEGER:
-                $stringifiedInt = is_scalar($data) && preg_match('#^[-+]?\d+$#', (string) $data) && ! is_float($data);
-                if (! is_int($data) && ! $stringifiedInt) {
+                if (is_string($data) || ! is_int($data)) {
```
scaytrase: same thoughts about redundancy here

src/PSR7/Validators/PathValidator.php
```diff
@@ -49,8 +49,9 @@ private function validateRequest(OperationAddress $addr, RequestInterface $messa
         $validator        = new SchemaValidator($this->detectValidationStrategy($message));
 
         foreach ($pathParsedParams as $name => $value) {
+            $parameter = RequestParameter::fromSpec($specs[$name]);
```
scaytrase: I think with new usages this class should have new name. `SerializedParameter` maybe?

src/Schema/Keywords/Type.php
m0003r: Fixed

src/PSR7/Validators/PathValidator.php
m0003r: Renamed it

m0003r:
Thank you for quick reply

I have implemented deserialization for headers (including multipart POST) and cookies, but not yet for urlencoded post params. There would be significant amount of duplication between `QueryArgumentsValidator` and yet not existing `BodyArgumentsValidator`. Also there is similar code in `RequestCookieValidator` and `HeaderValidator`. May be a some new class should be introduced, taking array of specs and array of string values?

scaytrase:
>  There would be significant amount of duplication

Well, we have two ways, get working code with duplications to discuss refactoring or just skip this step if you have a plan already in your mind how it should look like

m0003r:
I have implemented deserialization also for POST parameters and added some tests.
If I understand correctly, `SerializedParameter` now can be also used to deserialize objects in [different styles](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#style-values), as mentioned in #2

src/PSR7/Validators/ArrayValidator.php
```diff
@@ -0,0 +1,63 @@
...
+     * @param mixed[] $params
+     *
+     * @throws InvalidParameter
+     * @throws RequiredParameterMissing
+     */
+    public function validateArray(OperationAddress $addr, array $params, int $validationStrategy) : void
+    {
+        foreach ($this->specs as $name => $spec) {
+            if ($spec->required && ! array_key_exists($name, $params)) {
+                throw RequiredParameterMissing::fromNameAndAddr($name, $addr);
```
scaytrase: Not really important but I generally prefer do a rethrow outside method with `$addr` instead of passing it to the method for only be used to construct an exception. This make further code readers think that some logic depends on `$addr` inside this method. This class look like low-level utility class and with moving `$addr` and PSR-7 specific exception could even live in `League\OpenAPIValidation\Schema` namespace

src/PSR7/Validators/CookiesValidator/ServerRequestCookieValidator.php
```diff
@@ -37,38 +37,14 @@ public function __construct(array $specs)
...
-                $validator->validate($cookieValue, $this->specs[$cookieName]->schema);
-            } catch (SchemaMismatch $e) {
-                throw InvalidCookies::becauseValueDoesNotMatchSchema($cookieName, $cookieValue, $addr, $e);
-            }
+        $validator = new ArrayValidator($this->specs);
+
+        try {
+            $validator->validateArray($addr, $message->getCookieParams(), $this->detectValidationStrategy($message));
+        } catch (RequiredParameterMissing $e) {
+            throw InvalidCookies::becauseOfMissingRequiredCookie($e->name(), $addr);
```
scaytrase: Actually, as far as I can see, you do the rethrow I mentioned here, so there is really no need to bound `ArrayValidtor` to PSR-7 stuff (`$addr`)

src/PSR7/Validators/PathValidator.php
```diff
@@ -42,18 +41,14 @@ public function validate(OperationAddress $addr, MessageInterface $message) : vo
...
-            } catch (SchemaMismatch $e) {
-                throw InvalidPath::becauseValueDoesNotMatchSchema($name, (string) $value, $addr, $e);
-            }
-        }
+
+        try {
+            $validator->validateArray($addr, $pathParsedParams, $this->detectValidationStrategy($message));
+        } catch (InvalidParameter $e) {
+            throw InvalidPath::becauseValueDoesNotMatchSchema($e->name(), $e->value(), $addr, $e->getPrevious());
+        } // RequiredParameterMissing will not be thrown, all parameters are checking in parseParams
```
scaytrase: Well in this case I suggest rethrow to a `\LogicException`. This will never happen, but if unfortunatelly do - lib users could notify us to fix problem somewhere

scaytrase: Without it any stat analyzer will think that this method will be able to throw `RequiredParameterMissing`

m0003r: Good point, fixed it

src/PSR7/Validators/ArrayValidator.php
m0003r: That's definitely a redundancy, I haven't noticed it

scaytrase:
Actually I think I can fix it by myself

scaytrase:
Thanks @m0003r, sorry for long review

m0003r:
@scaytrase Thanks for the great library. I also have thoughts about XML validation (https://swagger.io/docs/specification/data-models/representing-xml/), maybe there will be another pull request sometime in the future :)

scaytrase:
Feel free to ping me if I forget to view it

PR #50, opened 2020-02-23
Fix path parameters parsing like {id}.{format}

m0003r:
OpenAPI paths like `/number/{id}.{format}` were converted to regex like `#/number/(?<id>[^/]+).(?<format>[^/]+)#` and due to unescaped dot it incorrectly parses string like `/number/1.json` to `['id' => '1.js', 'format' => 'n']`.

I have fixed such behavior with `preg_quote`. Also to distinguish parameter names from other parts braces balance is now validated. There is also test for it.


scaytrase:
Great work, @m0003r , thank you!

PR #55, opened 2020-04-23
Tests, which use the *Of keyword in combination with references won't…

rejinka:
… fail anymore.

League\OpenAPIValidation\Tests\Schema\SchemaValidatorTest::loadRawSchema() resolves the references exaclty like loadSchema() and the factories do.


scaytrase:
Thanks @rejinka !

PR #57, opened 2020-05-16
Expose required field violation cause

### LINKED ISSUE #14: Get the name of the invalid property
See https://github.com/lezhnev74/openapi-psr7-validator/issues/50 for discussion and details

scaytrase:
Introduce more specific RequiredMismatch extending KeywordMismatch which provides details of the field being missed

Fixes #14 

@coatesap
@canvural


canvural:
This is great, thank you!

Could this be applied in a more general way, or should be applied to all exceptions individually? And maybe `$e->dataBreadCrumb()->buildChain()` can be used to get the nested field names, in some cases.

scaytrase:
Yea, updating the chain sound better idea, let me try

scaytrase:
@canvural I've updated the patch, check it please

canvural:
@scaytrase Looks good!

Issue #60, opened 2020-05-18
Empty Content-Type generate a TypeError


hpatoio:
When we try to validate a request with `Content-Type` set to empty string the system throw the error: 

```
Return value of League\OpenAPIValidation\PSR7\Validators\BodyValidator\BodyValidator::messageContentType() must be of the type string or null, boolean returned
 ```

This happens because the function `strtok` can return a boolean. 

> This function may return Boolean FALSE, but may also return a non-Boolean value which evaluates to FALSE. Please read the section on Booleans for more information. Use the === operator for testing the return value of this function.

See https://www.php.net/manual/en/function.strtok.php#refsect1-function.strtok-notes for details.

PR #61, opened 2020-05-18
Fix TypeError in case of empty string a Content-Type

hpatoio:
See #60 for details


scaytrase:
Nice catch! Can you please apply `vendor/bin/phpcbf` in order I can merge this?

hpatoio:
I setup the hook as described in the doc ... but there were some errors because I'm using a docker. Fixing manually.

hpatoio:
Pushed CS fixes. If is ok I can squash and merge.

scaytrase:
Thanks @hpatoio !

PR #63, opened 2020-05-24
Respect uploaded file tree instead of flat array

### LINKED ISSUE #62: Multipart form multiple file uploads
I seem to have run into an issue that seems to go against what I can find in online documentation.

According to the documentation on [swagger's website](https://swagger.io/docs/specification/describing-request-body/file-upload/) on multiple file uploads, you need to specify a parameter as an array of binary strings.

However, this package says that is incorrect.

Instead, I get `Body does not match schema for content-type \"multipart\/form-data\" for Request [put \/clam\/scan]`:

image

However, if I set it to a singular binary string (not an array), it works with no issue:

image

I used exactly the same syntax that is provided in that documentation, except with the name "upload" instead of filename.

I'm not sure if swagger is wrong here, or if this validator is, but one of them is incorrect.

For my project, it's not a big deal, but something worth mentioning.

scaytrase:
Fixes #62 

cc @m50

### DIFF
--- a/src/PSR7/Validators/BodyValidator/MultipartValidator.php
+++ b/src/PSR7/Validators/BodyValidator/MultipartValidator.php
@@ -9,6 +9,7 @@ use cebe\openapi\spec\Header;
 use cebe\openapi\spec\MediaType;
 use cebe\openapi\spec\Schema;
 use cebe\openapi\spec\Type as CebeType;
+use InvalidArgumentException;
 use League\OpenAPIValidation\PSR7\Exception\NoPath;
 use League\OpenAPIValidation\PSR7\Exception\Validation\InvalidBody;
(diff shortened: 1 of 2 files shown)


scaytrase:
@m50 can you please verify this PR fixes your problem before merge?

m50:
Hi @scaytrase,

That works now! It does require that the name is `upload[]` which isn't exactly as it's mentioned in the docs I had provided, but it works more similarly to how things work in PHP, so I'll accept it.

Thank you!

scaytrase:
There is nothing I can do about `upload` or `upload[]` since it's php internals I hope. Thanks for checking

scaytrase:
Released as 0.8.2

PR #65, opened 2020-06-09
Allow about:blank as uri

canvural:
It is an edge case. But I think `about:blank` should also be a valid URI. For example, [this](https://tools.ietf.org/html/rfc7807#section-3.1) RFC uses `about:blank` as default value when no URI given.

I encountered this when I was working with Twitter's OpenAPI spec. They are implementing the `problems detail` RFC and make use of `about:blank` in their spec.

### DIFF
--- a/src/Schema/TypeFormats/StringURI.php
+++ b/src/Schema/TypeFormats/StringURI.php
@@ -11,6 +11,10 @@ class StringURI
 {
     public function __invoke(string $value) : bool
     {
+        if ($value === 'about:blank') {
+            return true;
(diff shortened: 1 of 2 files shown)


scaytrase:
I'm wondering if any software further is able to deal this kind of urls? Like maybe any PSR stuff or guzzle or symfony components and so on. I'm wondering what happens if I try to fetch this URL programmaticalyy. 

Looks pretty valid. though, according to specs

scaytrase:
thanks @canvural !

Issue #32, opened 2019-10-31
Exception InvalidBody will not thrown (at 0.3.3/0.4.0)


Nobody1809:
Since version 0.3.3 and 0.4.0 the validator will not throw an InvalidBody Exception. 

Version 0.3.2 throws this validation exception.

scaytrase:
Could you please provide more details on which schema and data you are trying to validate?

Nobody1809:
I can give you just some code snippets:

schema.yaml:
`components:
  schemas:
    prescription:
      properties:
        exampleTyp:
          type: string
          enum:
            - VALID
            - STILLVALID`

using functions:
ValidatorBuilder -> fromYamlFile and ServerRequestValidator -> validate

until test data:
`$data = [
            'exampleTyp' => 'INVALID',
        ];

$this->expectException(InvalidBody::class);`

Output: (for version 0.3.3 and 0.4.0) 
Failed asserting that exception of type "OpenAPIValidation\PSR7\Exception\Validation\InvalidBody" is thrown.

scaytrase:
Components are just re-usable parts of schema,actually there is no schema defined in your sample. Consider re-check your actual schema/

We already have the test for case you are talking about

https://github.com/thephpleague/openapi-psr7-validator/blob/5d2a74ed7b94511c89b5c79a00835b72844fd012/tests/Schema/Keywords/EnumTest.php#L30-L48

Nobody1809:
I wrote a short example test for you. Please try this yaml schema and the used methods.

```php
public function testValidateFromYamlContentWithInvalideBody(): void
    {
        $filePath = __DIR__.'/../fixtures/schema.yaml';
        $contentYaml = file_get_contents($filePath);

        $url = '/api/test/create';
        $headers = new Headers();
        $headers->addHeaders(['Content-Type' => 'application/json']);

        $data = [
            'exampleTyp' => 'INVALID',
        ];

        $request = (new Request())->setMethod(Request::METHOD_POST)
                                      ->setUri($url)
                                      ->setHeaders($headers)
                                      ->setContent(json_encode($data))
                                      ->setPost(new Parameters([]));

        $serverRequest = Psr7ServerRequest::fromZend($request);

        $serverRequestValidator = (new ValidatorBuilder())->fromYaml($contentYaml)->getServerRequestValidator();
        $this->expectException(InvalidBody::class);
        $serverRequestValidator->validate($serverRequest);
    }
```
```yaml
[...]

PR #66, opened 2020-06-10
Infer object type if properties is set

canvural:
Another possible edge case. Sometimes` type: object` might not be in the schema. This usually happens if `properties` is combined with `*Of` Some simple example is:

```JSON
{
  "oneOf": [
    {
      "type": "object",
      "properties": {
        "bar": {
          "type": "string"
        }
      }
    }
  ],
  "properties": {
    "foo": {
      "type": "string"
    }
  }
}
```

This schema should fail the validation for the following data:
```JSON
{
	"foo": 123
}
```

because `foo` is not a string.

This PR adds ability to guess the object type by checking if `properties` is set on the schema.

### DIFF
--- a/src/Schema/SchemaValidator.php
+++ b/src/Schema/SchemaValidator.php
@@ -6,6 +6,7 @@ namespace League\OpenAPIValidation\Schema;
 
 use cebe\openapi\spec\Schema as CebeSchema;
 use cebe\openapi\spec\Type as CebeType;
+use League\OpenAPIValidation\Foundation\ArrayHelper;
 use League\OpenAPIValidation\Schema\Exception\SchemaMismatch;
 use League\OpenAPIValidation\Schema\Keywords\AllOf;
 use League\OpenAPIValidation\Schema\Keywords\AnyOf;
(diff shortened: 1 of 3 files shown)


scaytrase:
Actually it's a fix for #32 
Could you please try to incorporate test case from #34 into this one and check if it helps?

src/Schema/SchemaValidator.php
```diff
@@ -126,7 +127,7 @@ public function validate($data, CebeSchema $schema, ?BreadCrumb $breadCrumb = nu
                 (new Items($schema, $this->validationStrategy, $breadCrumb))->validate($data, $schema->items);
             }
 
-            if ($schema->type === CebeType::OBJECT) {
+            if ($schema->type === CebeType::OBJECT || (isset($schema->properties) && is_array($data))) {
```
scaytrase, line 130: I hope we should check for assoc arrays here. There is an util for this 
https://github.com/thephpleague/openapi-psr7-validator/blob/master/src/Foundation/ArrayHelper.php#L23

canvural: Added :+1:

canvural:
> Actually it's a fix for #32
> Could you please try to incorporate test case from #34 into this one and check if it helps?

I tried the failing test case from #34 and looks like this PR fixes that :+1:

scaytrase:
Thanks @canvural

PR #70, opened 2020-07-24
Add new exceptions based on message type

lezhnev74:
- exceptions are in '\League\OpenAPIValidation\PSR7\Exception\*'
- each validator now throws only this exceptions

### DIFF
--- a/src/PSR7/Exception/InvalidRequestMessage.php
+++ b/src/PSR7/Exception/InvalidRequestMessage.php
@@ -6,7 +6,7 @@ namespace League\OpenAPIValidation\PSR7\Exception;
 
 class InvalidRequestMessage extends ValidationFailed
 {
-    public static function fromOriginal(ValidationFailed $e) : self
+    public static function because(ValidationFailed $e) : self
     {
         return new static('Request message failed validation', 0, $e);
     }
--- a/src/PSR7/Exception/InvalidRequestMessage.php
+++ b/src/PSR7/Exception/InvalidRequestMessage.php
@@ -4,12 +4,9 @@ declare(strict_types=1);
 
 namespace League\OpenAPIValidation\PSR7\Exception;
 
-use Exception;
-use Throwable;
-
-class InvalidRequestMessage extends Exception
+class InvalidRequestMessage extends ValidationFailed
 {
-    public static function fromOriginal(Throwable $e) : self
+    public static function fromOriginal(ValidationFailed $e) : self
     {
         return new static('Request message failed validation', 0, $e);
     }
--- a/src/PSR7/Exception/InvalidRequestMessage.php
+++ b/src/PSR7/Exception/InvalidRequestMessage.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\OpenAPIValidation\PSR7\Exception;
+
+use Exception;
+use Throwable;
+
+class InvalidRequestMessage extends Exception
+{
+    public static function fromOriginal(Throwable $e) : self
+    {
+        return new static('Request message failed validation', 0, $e);
+    }
+}
--- a/src/PSR7/Exception/InvalidServerRequestMessage.php
+++ b/src/PSR7/Exception/InvalidServerRequestMessage.php
@@ -4,12 +4,9 @@ declare(strict_types=1);
 
 namespace League\OpenAPIValidation\PSR7\Exception;
 
-use Exception;
-use Throwable;
-
-class InvalidServerRequestMessage extends Exception
+class InvalidServerRequestMessage extends ValidationFailed
 {
-    public static function fromOriginal(Throwable $e) : self
+    public static function fromOriginal(ValidationFailed $e) : self
     {
         return new static('Server Request message failed validation', 0, $e);
     }
--- a/src/PSR7/Exception/InvalidServerRequestMessage.php
+++ b/src/PSR7/Exception/InvalidServerRequestMessage.php
@@ -6,7 +6,7 @@ namespace League\OpenAPIValidation\PSR7\Exception;
 
 class InvalidServerRequestMessage extends ValidationFailed
 {
-    public static function fromOriginal(ValidationFailed $e) : self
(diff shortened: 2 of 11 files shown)


src/PSR7/Exception/InvalidServerRequestMessage.php
```diff
@@ -0,0 +1,16 @@
...
+declare(strict_types=1);
+
+namespace League\OpenAPIValidation\PSR7\Exception;
+
+use Exception;
+use Throwable;
+
+class InvalidServerRequestMessage extends Exception
+{
+    public static function fromOriginal(Throwable $e) : self
```
scaytrase, line 12: I'm not sure whether `Throwable` is not overkill there. Since we do a rethrow, previous `ValidationFailed` would cover all the cases and will be consised with the message - not any throwable is failed validation

src/PSR7/ResponseValidator.php
```diff
@@ -38,9 +40,13 @@ public function getSchema() : OpenApi
...
-            new ResponseAddress($opAddr->path(), $opAddr->method(), $response->getStatusCode()),
-            $response
-        );
+        try {
+            $this->validator->validate(
+                new ResponseAddress($opAddr->path(), $opAddr->method(), $response->getStatusCode()),
+                $response
+            );
+        } catch (Throwable $e) {
+            throw InvalidResponseMessage::fromOriginal($e);
```
scaytrase, line 49: the phpdoc is now incorrect, it should declare other throw

src/PSR7/Exception/InvalidServerRequestMessage.php
lezhnev74: Yeah, true. The current way completely hides ANY errors possible: some runtime problems, actual validation problems, syntax errors, god knows what else can go wrong in there. I don't mind if we change this and catch `ValidationError` only.

src/PSR7/Exception/InvalidRequestMessage.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\OpenAPIValidation\PSR7\Exception;
+
+use Exception;
+use Throwable;
+
+class InvalidRequestMessage extends Exception
```
lezhnev74, line 10: Btw, @scaytrase should we extend from ValidationFailed as well here?

scaytrase: Actually we can. Do not see any problems with it now

src/PSR7/Exception/InvalidMessage.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        return $this->code === self::MESSAGE_RESPONSE;
+    }
+
+    public function isRequest() : bool
+    {
+        return $this->code === self::MESSAGE_REQUEST;
+    }
+
+    public function isServerRequest() : bool
```
scaytrase, line 40: I feel previous approach with separate exceptions way more extensible since anyone can write own validator and add own exception similar to existing. with this checks I don't see it's possible

scaytrase: They can still extend ValidationFailed

lezhnev74: Yeap, it makes sense! let me revert it

lezhnev74:
well, the only thing to fix is:
image

scaytrase:
Easy fix will be just unwrap one exception layer.

lezhnev74:
do you want to show how and push a patch?

scaytrase:
If no one picks it in 2-3 hours then yes, I can

sunkan:
Hi

This pull-request break any custom middlewares that catches the more specific exceptions thrown by the validators.

Why make the change inside the PSR7 validators when the original problem was related to the PSR15 `ValidationMiddleware`

If I uses just one of the validators I know that the exception throw from it's connected to that one and making me unwrap the exception is just unnecessary 

In my eyes this would have been a much cleaner soultion
```
try {
    // 1. Validate request
    $matchedOASOperation = $this->requestValidator->validate($request);
} catch(ValidationFailed $e) {
    throw InvalidServerRequestMessage::because($e);
}

// 2. Process request
$response = $handler->handle($request);

try {        
    // 3. Validate response
    $this->responseValidator->validate($matchedOASOperation, $response);
} catch(ValidationFailed $e) {
    throw InvalidResponseMessage::because($e);
}
return $response;
```

scaytrase:
Sounds reasonable but 0.x versions should be always considered unstable

sunkan:
Yes sure that it unstable.

But my point still stands that you shouldn't wrap the exception inside the validators because it doesn't make sense.
Why force the developer that directly use a validator to unwrap the exceptions when they already now that the exception comes from it comes from it. And it breaks the concept of being allowed to catch more specific errors from the validator

scaytrase:
Actually I'm totally not sure whether this package should provide customs like slip middleware at all.

sunkan:
That's also fine.

My problem is with the wrapping of all exceptions in the validators makes the validators less useful. Because I'll need to unwrap the exception to be able to use it.

scaytrase:
@lezhnev74 what do you think? I see @sunkan reasons legit, but enlarging frameworks support (with special exceptions pack) in this lib seems odd to me, I'd rather suggest drop it at all

lezhnev74:
I understand what @sunkan suggests, when you use validators directly you don't need the extra layer of exceptions. It only makes sense in the situation when you validate multiple messages and you need to understand which one failed (exactly what middleware does). So let me actually follow @sunkan suggestion and move those exceptions to PSR-15 namespace.

lezhnev74:
@scaytrase if you are ok with this, then please do merge

lezhnev74:
@scaytrase btw, speaking of middleware, I included it in the package in the very beginning because I understand that this package's major usecase is a middleware. It is completely optional, but this is a care after developers who will use this. Does not cost much to ship with the package but helps a lot. However I understand your concerns about the "framework specific" tools... I wouldn't add any new framework specific stuff in this in future, thats true.

PR #76, opened 2020-08-03
Feature/handle path parameters

m8nky:
Taking over PR #40 , wanna do another try. Reviews welcome.


src/PSR7/SpecFinder.php
```diff
@@ -37,6 +38,39 @@ public function __construct(OpenApi $openApi)
         $this->openApi = $openApi;
     }
 
+    /**
+     * @return Parameter[]
+     *
+     * @throws NoPath
+     */
+    public function findPathSpecs(OperationAddress $addr) : array
```
scaytrase, line 46: I suggest to tune naming since I was confused for couple of minutes figuring out it's not a recursion method. Maybe something like `findAllPathSpecs` or `findPathAndOperationSpecs`

m8nky: fixed

src/PSR7/SpecFinder.php
```diff
@@ -74,37 +108,49 @@ public function findPathSpec(OperationAddress $addr) : PathItem
         return $pathSpec;
     }
 
-    /**
-     * @return Parameter[]
-     *
-     * @throws NoPath
-     */
-    public function findPathSpecs(OperationAddress $addr) : array
+    private function findPathSimple(OperationAddress $addr) : ?PathItem
```
scaytrase, line 111: I'll keep some comments from previous PR

I'd rather suggest to name these new methods something like findExactPath and findMatchingTemplatePath or something like this. Current naming gives no real understanding what each method does under the hood.

src/PSR7/SpecFinder.php
```diff
@@ -74,37 +108,49 @@ public function findPathSpec(OperationAddress $addr) : PathItem
...
+        $results = $finder->search();
+        if (count($results) === 1) {
+            return $this->findPathSimple($results[0]);
+        }
 
-        foreach ($spec->parameters as $p) {
-            if ($p->in !== 'path') {
-                continue;
-            }
+        return null;
```
scaytrase, line 124: Another problem from original PR. Returning null from here is processed with `NoPath` exception while it could be more than one matching

src/PSR7/SpecFinder.php
```diff
@@ -74,37 +108,49 @@ public function findPathSpec(OperationAddress $addr) : PathItem
...
+    }
 
-            $pathSpecs[$p->name] = $p;
+    /**
+     * @throws NoCallback
+     */
+    private function findCallbackInOperation(CallbackAddress $addr, Operation $operation) : Operation
```
scaytrase, line 130: minor but I suggest to keep current public-protected-private order. this make PR easier even if not reflected in automatic style guides

src/PSR7/SpecFinder.php
m8nky: Agreed, what's your suggestion.
1. Ambiguous exception - in case of count>1
2. Return first match.
3. Apply a ruleset to determine, which spec might be the right one?

Maybe we start with (2) and think about a more mature solution in upcoming releases.

src/PSR7/SpecFinder.php
m8nky: done

src/PSR7/SpecFinder.php
m8nky: Solution 2 for now - return first match

src/PSR7/SpecFinder.php
m8nky: done. Passed the responsibility of determining the path matching method to PathFinder.

m8nky:
- [ ] Mixed methods for same path might lead to wrong spec chosen. Still fixing.
- [ ] Implement test, that partial matches lead to the correct spec.

scaytrase:
Sorry for delay, was a bit busy

scaytrase:
LGTM 👍 Great work @m8nky, thanks for picking and making this done!

b-boehne:
Thanks @m8nky for picking this up!

PR #78, opened 2020-08-20
Allow PHPUnit 9.x in composer.json

thunderer:


scaytrase:
I don't think supporting 3 major versions of phpunit it worth it. We can't start using phpunit 9 features why we supporting older phpunit for older php verions. What the purpose of allowing phpunit9?

thunderer:
@scaytrase I'd like to use this library in a project which uses PHPUnit 9. Since tests pass on all PHPUnit versions we can allow it without blocking projects who would otherwise need to downgrade their PHPUnit versions. The library doesn't need to use any PHPUnit 9 specific features.

scaytrase:
phpunit is in `require-dev` section, so it doesn't affect any dependants.

thunderer:
I upgraded the library manually as `composer update` for some reason didn't pick up newer version for `^0.5` constraint and yes, you're right. I'm closing this PR then.

scaytrase:
`0.x` works a bit different with constraints

https://jubianchi.github.io/semver-check/#/^0.5/0.12

PR #80, opened 2020-09-08
[Hotfix][Issue-79] Avoid serialization when resolving servers

### LINKED ISSUE #79: Bad performance in the PathFinder
Hello,
First thanks for the great work you've done here.

I wanted to use this library to validate the requests/responses of our API within our codeception tests.
During my experimentation with this project I noticed a performance issue related to the `PathFinder` service.
I can't share my `swagger.yaml` but I did my tests with a 2MB file.

The issue is that searching a path is too long, this is because of the `PathFinder::findServersForOperation()` method.
More specifically the instruction `$this->openApiSpec->getSerializableData()` which takes in average 0.8s.

The method `PathFinder::search()` which trigger the previous calls is used in multiple places which result in an average time of 4s to validate both the request and the response.

Possible solution:
When investigating the issue I realized that this part seems unnecessary:
```php
        // 3. Check servers on root level
        if (array_key_exists('servers', (array) $this->openApiSpec->getSerializableData()) &&
            count($this->openApiSpec->servers) > 0) {
            return $this->openApiSpec->servers;
        }

        // fallback
        return [new Server(['url' => '/'])];
```

Indeed in the `OpenApi` class, the `servers` property is declared with a default value of `[new Server(['url' => '/'])]`, just as the fallback you used.
Therefore there is no need to test if the key actually exists neither as to ensure there is at least 1 server defined and this could simply be:
```php
        return $this->openApiSpec->servers;
```

I'm not familiar with the OAS3 specification nor as I am with the `cebe/php-openapi` library, so I didn't want to open a PR without discussing it here first.
I don't think my proposition will have any impact since the `SpecBaseObject::getSerializableData()` method does not affect the state of the object.

elythyr:
Fix #79 

Let me know if you think of other scenario to test or if I should include the tests to the `PathFinderTest` class directly.

I have split my commits by habit but don't hesitate to ask if you want me to squash them.

### DIFF
--- a/src/PSR7/PathFinder.php
+++ b/src/PSR7/PathFinder.php
@@ -8,7 +8,6 @@ use cebe\openapi\spec\OpenApi;
 use cebe\openapi\spec\PathItem;
 use cebe\openapi\spec\Server;
 use const PHP_URL_PATH;
-use function array_key_exists;
 use function count;
 use function ltrim;
 use function parse_url;
@@ -154,24 +153,16 @@ class PathFinder
         $operation = $path->getOperations()[$opAddress->method()];
 
         // 1. Check servers on operation level
-        if (array_key_exists('servers', (array) $operation->getSerializableData()) &&
-            count($operation->servers) > 0) {
+        if (isset($operation->servers) && count($operation->servers) > 0) {
             return $operation->servers;
         }
 
         // 2. Check servers on path level
-        if (array_key_exists('servers', (array) $path->getSerializableData()) &&
-            count($path->servers) > 0) {
+        if (isset($path->servers) && count($path->servers) > 0) {
             return $path->servers;
         }
 
-        // 3. Check servers on root level
-        if (array_key_exists('servers', (array) $this->openApiSpec->getSerializableData()) &&
-            count($this->openApiSpec->servers) > 0) {
-            return $this->openApiSpec->servers;
-        }
-
-        // fallback
-        return [new Server(['url' => '/'])];
+        // 3. Fallback with servers on root level
+        return $this->openApiSpec->servers;
     }
 }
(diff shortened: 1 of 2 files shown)


elythyr:
Scrutinizer and Travis both failed because of `PHP Parse error:  syntax error, unexpected end of file in /home/scrutinizer/build/tests/FromCommunity/Issue79Test.php on line 198`
Did you already have this error, I checked the oher files there is no empty line at the end.

For the second failure in Travis it can't be reproduce locally, I started `phpcs` manually and enabled the git hook anyway.
Plus it seems to came from an error in the vendor package directly...

scaytrase:
I think unexpected EOF is not about missing or existing EOL, but for something like uncolsed `{`. Let me chech

scaytrase:
I've got a problem (PHPStorm doesn't complain about it thought). Having HEREDOC closure tag cannot have comma after it

i.e
```
<<<SPEC
data
SPEC,
```

comma is not being read.

scaytrase:
Dirty fix - move specs to variables

elythyr:
Done, nice catch. Vim was not complaining neither as PHP but it might be because I'm running in 7.4

scaytrase:
LGTM 👍

scaytrase (approved):

lezhnev74 (approved):

scaytrase:
Thanks @elythyr! Great research and fix

elythyr:
Thanks for the reactivity!
Is a tag 0.12.1 is planed ?

scaytrase:
Yea, I was looking what else could be picked up to the release

scaytrase:
https://github.com/thephpleague/openapi-psr7-validator/releases/tag/0.12.1

PR #81, opened 2020-09-09
add pattern validation

foritel:
Allow set pattern with modifiers. Only `#` delimiter support with modifiers.

### DIFF
--- a/src/Schema/Keywords/Pattern.php
+++ b/src/Schema/Keywords/Pattern.php
@@ -36,7 +36,9 @@ class Pattern extends BaseKeyword
             throw InvalidSchema::becauseDefensiveSchemaValidationFailed($e);
         }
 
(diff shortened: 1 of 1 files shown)


lezhnev74:
Good suggestion! I think this sort of things was already discussed in https://github.com/thephpleague/openapi-psr7-validator/pull/64#issuecomment-643139137

The current design was done intentionally: the pattern should be interpreted as regular expression body (no tags or modifiers should be included).

foritel:
> Good suggestion! I think this sort of things was already discussed in [#64 (comment)](https://github.com/thephpleague/openapi-psr7-validator/pull/64#issuecomment-643139137)
> 
> The current design was done intentionally: the pattern should be interpreted as regular expression body (no tags or modifiers should be included).
> 
> p.s. it is a good practice to add tests with the PR for the code to prove it works.

Thanks for answer. Didn't see the discussion, I’ll go check it out.

PR #82, opened 2020-09-09
Pattern utf-8 support

foritel:
Hello, can you add UTF-8 support to your project? Since this implementation is in php, and UTF-8 is used quite often, I think you can do this :)

### DIFF
--- a/src/Schema/Keywords/Pattern.php
+++ b/src/Schema/Keywords/Pattern.php
@@ -24,6 +24,7 @@ class Pattern extends BaseKeyword
      * not implicitly anchored.
      *
      * @param mixed $data
+     * @param string $pattern
      *
      * @throws KeywordMismatch
      */
--- a/src/Schema/Keywords/Pattern.php
+++ b/src/Schema/Keywords/Pattern.php
@@ -38,6 +38,13 @@ class Pattern extends BaseKeyword
 
         $pattern = sprintf('#%s#', str_replace('#', '\#', $pattern));
 
+        if (
+            function_exists('mb_check_encoding') &&
+            (mb_check_encoding($data, 'UTF-8') || mb_check_encoding($pattern, 'UTF-8'))
+        ) {
+            $pattern .= 'u';
+        }
+
         if (! preg_match($pattern, $data)) {
             throw KeywordMismatch::fromKeyword('pattern', $data, sprintf('Data does not match pattern \'%s\'', $pattern));
         }
--- a/src/Schema/Keywords/Pattern.php
+++ b/src/Schema/Keywords/Pattern.php
@@ -8,6 +8,8 @@ use League\OpenAPIValidation\Schema\Exception\InvalidSchema;
 use League\OpenAPIValidation\Schema\Exception\KeywordMismatch;
 use Respect\Validation\Exceptions\ExceptionInterface;
 use Respect\Validation\Validator;
+use function function_exists;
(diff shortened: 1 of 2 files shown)


src/Schema/Keywords/Pattern.php
```diff
@@ -38,6 +39,13 @@ public function validate($data, string $pattern) : void
 
         $pattern = sprintf('#%s#', str_replace('#', '\#', $pattern));
 
+        if (
+            function_exists('mb_check_encoding') &&
+            (mb_check_encoding($data, 'UTF-8') || mb_check_encoding($pattern, 'UTF-8'))
```
scaytrase, line 44: should be both data AND pattern be valid UTF-8 in order to use `u` ?

tests/Schema/Keywords/PatternTest.php
```diff
@@ -21,7 +21,8 @@ public function validDataProvider() : array
             ['foof', 'foof'], // Tests adding anchors when first and last character is same
             ['1foo1', '1foo1'], // Tests adding anchors when first and last character is same with numbers
             ['^#\d+$', '#123'], // Tests adding anchors to string which has #
-            ['^#(\d+)#$', '#123#'], // Tests adding anchors to string which has multiple#
+            ['^#(\d+)#$', '#123#'], // Tests adding anchors to string which has multiple #
+            ['^[\pL\pNd\s\-]*$', 'Строка в UTF-8'], // Tests adding anchors and work with "u" modifier
```
scaytrase, line 25: I think one should add more examples not only with special patterns but with things like `['лаб', 'блабла']`. Also I think we need opposite example, where string is not a valid UTF-8 string

src/Schema/Keywords/Pattern.php
cebe: I think if the pattern is invalid UTF-8 that should result in an error instead of siliently not adding the modifier.
If data is not valid UTF-8 it would simply not match the condition, so it should not have any influence on the modifier.

scaytrase: > I think if the pattern is invalid UTF-8 that should result in an error

sounds valid. if I understand the spec correctly - pattern should be always valid UTF-8 string. So we can throw schema validation failed here defensively

scaytrase (changes requested):
Changes for the condition required

src/Schema/Keywords/Pattern.php
foritel: Yes, for the "u" modifier to work, both values must be valid UTF-8, but this is the responsibility of the developer, not the validator, as it seems to me. With this check, we only achieve the expected behavior of the validator when one of the values contains UTF-8 characters.

tests/Schema/Keywords/PatternTest.php
foritel: You're right, I'll add additional examples)

src/Schema/Keywords/Pattern.php
cebe: 1. Data sent in the request should not influence the pattern that is going to validate that data. Sending invalid UTF-8 will allow someone to influence the validation condition. This should not be allowed.
2. Invalid UTF-8 pattern is an error in the specification and should be handled like that.

scaytrase:
No updates for two months. Feel free to reopen it, when it's ready to be finished

Issue #89, opened 2020-11-02
Valid URIs are rejected if they are not a URL


garethellis36:
The class `StringURI` uses `filter_var($value, FILTER_VALIDATE_URL) !== false` to validate properties of type `string` and format `uri`. This means that a value which is a valid URI but not a URL (e.g. `auth:invalid`) fails validation.

I think the `StringURI` could be updated to use the `league/uri` package instead. Would you be open to a PR for this?

scaytrase:
Sure, at least I'd like to see PR with some failing tests in order to better understand the problem

PR #90, opened 2020-11-02
#89 StringURI now handles validation of valid URIs which aren't URLs

garethellis36:
At the moment this won't pass build, the first commit only contains failing tests. I copied the list of valid and invalid URIs from the `league/uri` package's tests. See: https://github.com/thephpleague/uri/blob/master/tests/UriStringTest.php

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
         "ext-json": "*",
         "cebe/php-openapi": "^1.3",
         "dflydev/fig-cookies": "^2.0",
+        "league/uri": "^6.3",
         "psr/cache": "^1.0",
         "psr/http-message": "^1.0",
         "psr/http-server-middleware": "^1.0",
--- a/src/Schema/TypeFormats/StringURI.php
+++ b/src/Schema/TypeFormats/StringURI.php
@@ -4,17 +4,18 @@ declare(strict_types=1);
 
 namespace League\OpenAPIValidation\Schema\TypeFormats;
 
-use const FILTER_VALIDATE_URL;
-use function filter_var;
+use League\Uri\UriString;
+use League\Uri\Exceptions\SyntaxError;
 
 class StringURI
 {
     public function __invoke(string $value) : bool
     {
-        if ($value === 'about:blank') {
+        try {
+            UriString::parse($value);
             return true;
+        } catch (SyntaxError $error) {
+            return false;
         }
-
-        return filter_var($value, FILTER_VALIDATE_URL) !== false;
     }
 }
--- a/src/Schema/TypeFormats/StringURI.php
+++ b/src/Schema/TypeFormats/StringURI.php
@@ -4,8 +4,8 @@ declare(strict_types=1);
 
 namespace League\OpenAPIValidation\Schema\TypeFormats;
 
-use League\Uri\UriString;
 use League\Uri\Exceptions\SyntaxError;
+use League\Uri\UriString;
 
 class StringURI
 {
@@ -13,6 +13,7 @@ class StringURI
     {
         try {
             UriString::parse($value);
+
             return true;
         } catch (SyntaxError $error) {
             return false;
--- a/tests/Schema/TypeFormats/StringURITest.php
+++ b/tests/Schema/TypeFormats/StringURITest.php
@@ -16,6 +16,9 @@ class StringURITest extends TestCase
     public function greenURIDataProvider() : array
     {
         return [
+            'about:blank' => [
+                'about:blank'
+            ],
             'scheme with non-leading digit' => [
                 's3://somebucket/somefile.txt',
             ],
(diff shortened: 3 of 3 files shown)


garethellis36:
Second commit adds `league/uri` package and uses that to validate the value instead of `filter_var()`

scaytrase:
Are all of these allowed according to OAS3 spec ?

garethellis36:
> Are all of these allowed according to OAS3 spec ?

According to [the Json Schema definition](https://json-schema.org/understanding-json-schema/reference/string.html#id13), when something is type `string` and format `uri`, it means `A universal resource identifier (URI), according to RFC3986.`

`league/uri` states that the package is RFC3986-compliant. So all of the valid URIs from its test suite should be valid in the context of an OpenAPI spec as well.

composer.json
```diff
@@ -25,6 +25,7 @@
         "ext-json": "*",
         "cebe/php-openapi": "^1.3",
         "dflydev/fig-cookies": "^2.0",
+        "league/uri": "^6.3",
```
garethellis36, line 28: I suppose the main problem with this approach is that it could cause BC issues for a user who is already dependent on a different version of `league/uri`.

scaytrase (approved):
LGTM 👍

garethellis36 (commented):
> LGTM 👍

Are you happy to add a dependency to the other package at that specific version? Won't it cause problems for someone who already has a dependency on (for example) `league/uri: "6.2"`?

I'm not really sure how libraries generally address this sort of thing. It looks like `6.0.0` has the same API, so it should be OK to change the dependency to `6.*`. But you couldn't have your library and also install `league/uri` at version 5, for example (the API is different there).

As an alternative, I could adapt the code from the URI library and re-implement it in this library without the additional dependency. The obvious downside is that there is then more code for you to maintain, and it would be harder to keep up with upstream changes if it emerged that there was an issue with the parser in the URI lib.

composer.json
scaytrase: Being dependend on `^6.1` for example is not a BC when requesting a lib bringing requirement of `^6.3`. It'will just make you to update to the `^6.3` (which comforms `^6.1`). The problems are then constraints are misused, like `~6.2.1` or `6.0.3` or whatever unnecessary strict. 

This should be resolved by the people who decided to make these constraints.

If people really struggle to bring `^6.3` to their project - downgrading dependency to `^6.x` is good until it's working as same as `^6.3` in terms of our usage. And such a downgrade is backward compatible

scaytrase:
> As an alternative, I could adapt the code from the URI library and re-implement it in this library without the additional dependency. The obvious downside is that there is then more code for you to maintain, and it would be harder to keep up with upstream changes if it emerged that there was an issue with the parser in the URI lib.

This is not how libraries should be used for sure. No one wants to maitain two libs with same code instead of one

composer.json
garethellis36: What about the case where someone already has a dependency on `5.1` and doesn't want to, or can't, upgrade to 6.x? When they try to upgrade the openapi validator package, composer will error because there will be a dependency conflict between 6.x (this package) and 5.x (their other dependency).

scaytrase: Supporting out of date dependencies is always the matter of additional resources. Why we are talking only about 5.x, not 4.x, 3.x, 2.x or people who's dependensies do not allow to bring the lib at all (imagine they have another dependency, which have `conflicts: {"league/uri":"*"}`)

In general we do support only maintained verions of dependecies (latest minor by default, according to semver). Any wider assumptions MAY be implemented, but with real use cases and better with resources from people who really need this

garethellis36: Sorry, maybe I wasn't explaining myself well :)

I am not suggesting that you should support out-of-date dependencies. If we want to use `league/uri` as a dependency, we absolutely should use the latest version. I just wanted to flag up that if there is someone who depends on `league/uri` in any previous version before the `6.x` line, when they come to update *this* library, they will get a composer error, and will be forced to update `league/uri` as well.

Anyway, do you want me to leave this as dependent on `^6.3` or change to something else? I'm happy to do whatever you'd prefer. I will also fix the PHPCS errors from the Travis build.

scaytrase: > when they come to update this library, they will get a composer error, and will be forced to update league/uri as well.

or they just stick to version of this lib they already have. no BC-breaks here.

> Anyway, do you want me to leave this as dependent on ^6.3 or change to something else? I'm happy to do whatever you'd prefer

`^6.3` is fine at the moment. We can downgrade the dependecy at any further moment easy.

garethellis36:
Hi @scaytrase, what else needs to happen before this PR can be merged? Thanks for your support :)

scaytrase:
Oh, sorry, thanks for reminding. I'll check it asap

scaytrase (approved):

Issue #25, opened 2019-10-19
Distinguish between assoc arrays and objects in validation messages


scaytrase:
See https://github.com/thephpleague/openapi-psr7-validator/issues/17#issuecomment-542760320

lezhnev74:
See #92 as well

dhaeckel:
I stumbled upon this problem while debugging a validation failure with a collegue. It took us longer than necessary, to find that we had this exact situation. I created a PR for providing a more suitable validation message and expanded the tests.

gordinskiy:
Is there something wrong with the proposed solutions?

Are there any plans to fix this?

dhaeckel:
@scaytrase @lezhnev74
Wanted to ping on this issue. If there is anything wrong with my solution let me know, I'm happy to change the PR.

mstyles:
Would love to see this fixed...

PR #92, opened 2020-11-03
Discuss a misleading message: "Value expected to be 'array', 'array' given."

lezhnev74:
Here is a problem found, see the diff.
The message is quite misleading, however it comes from the fact that "objects" and "arrays" in json are represented as just "arrays" in PHP... 
what do we do with this?

### DIFF
--- a/tests/Schema/Keywords/TypeTest.php
+++ b/tests/Schema/Keywords/TypeTest.php
@@ -20,6 +20,7 @@ final class TypeTest extends SchemaValidatorTest
             ['string', null, 'string value'],
             ['object', null, ['a' => 1]],
             ['array', null, ['a', 'b']],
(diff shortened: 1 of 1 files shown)


tests/Schema/Keywords/TypeTest.php
```diff
@@ -20,6 +20,7 @@ public function validDataProvider() : array
             ['string', null, 'string value'],
             ['object', null, ['a' => 1]],
             ['array', null, ['a', 'b']],
+            ['array', null, ['0' => 'a', '2' => 'b']], // this triggers: "Value expected to be 'array', 'array' given."
```
scaytrase, line 23: actually `object given` should be raised here. but we have an issue about this already, I hope

scaytrase: yea, this one https://github.com/thephpleague/openapi-psr7-validator/issues/25

lezhnev74:
already a discussion in #25

Issue #93, opened 2020-11-11
Unable to Differentiate Between Bad Paths and Bad Methods on Good Paths


jimdelois:
Given something like the following (the rethrown exceptions are part of the local framework):
```
        try {
            $this->requestValidator->validate($request);
        } catch (Exception\NoOperation $e) {
            throw new HttpException\HttpMethodNotAllowedException($request, 'Method Not Allowed', $e);
        } catch (Exception\NoPath $e) {
            throw new HttpException\HttpNotFoundException($request, 'Not Found', $e);
        }
```
... It seems that whatever I catch first catches everything.  That is to say, a bad path on the case above yields a `405` (unexpected), as does a bad method (expected).  If I reverse the two catches, the opposite occurs...  A bad path on the case yields a `404` (expected) and a bad method yields a `405` (unexpected).

The only way that's possible, based on one exception extending the other, is if the `NoOperation` is the only thing thrown in all cases.

From your own doc:
```
\League\OpenAPIValidation\PSR7\Exception\NoPath - path is not found in the spec
\League\OpenAPIValidation\PSR7\Exception\NoOperation - operation os not found in the path
```
I have not yet dug into the internals, but the documentation very strongly suggests that these are different outcomes, which would be in alignment with the HTTP spec.  Evidently they are not.

It seems a pretty significant issue that one cannot differentiate something that should be bubbled as a `404` vs. a `405`.  There also seems to be something that isn't working with the `NoContentType` that I can't possible trigger in tests.

Can someone educate me on how to work with these different exceptions such that I can appropriately recast them up to the client?  Is it merely a documentation issue that I'm not seeing?  If I've overlooked something, can anyone provide some input?

Any assistance is much appreciated!!  I'm assuming there's a shortcoming with myself, but if there's actually something to be improved within the library then perhaps I could eventually attempt to patch it.

scaytrase:
Hello, can you please open PR with a test with failure example? It seems I can't find any test that utilize no operation exception

jimdelois:
Sure thing - here you go: https://github.com/thephpleague/openapi-psr7-validator/pull/94

Because of the inheritance and the fact the `NoOperation` is thrown in both cases, I can't use a multiple try/catch blocks to determine if it was a bad path or a bad operation and raise accordingly (as described in the original filing).

As an aside, the Contribution notes in the README are outdated.

scaytrase:
Thanks. I'll take a look on it

[...]

PR #94, opened 2020-11-11
Adding Example Test Case for Issue #93

jimdelois:
This is an example test case to reproduce the oddity noted in #93 where we have no ability to actually catch a `NoPath` Exception separately from a `NoOperation` Exception because both outcomes throw the latter.  This means one cannot bubble them separately as, e.g., a `404` and a `405` respectively.


scaytrase:
I've incorporated your test sampe into the fix, so we can close this. Thank you!

PR #95, opened 2020-11-21
Jimdelois bugfix/issue 93

### LINKED ISSUE #93: Unable to Differentiate Between Bad Paths and Bad Methods on Good Paths

scaytrase:
Dirty fix for #93. I think it's decreases performance, so we need to find better way to handle that, but just to start discussion

Fixes #93  

cc. @jimdelois

### DIFF
--- a/src/PSR7/RequestValidator.php
+++ b/src/PSR7/RequestValidator.php
@@ -7,6 +7,7 @@ namespace League\OpenAPIValidation\PSR7;
 use cebe\openapi\spec\OpenApi;
 use League\OpenAPIValidation\PSR7\Exception\MultipleOperationsMismatchForRequest;
 use League\OpenAPIValidation\PSR7\Exception\NoOperation;
+use League\OpenAPIValidation\PSR7\Exception\NoPath;
 use League\OpenAPIValidation\PSR7\Exception\ValidationFailed;
 use League\OpenAPIValidation\PSR7\Validators\BodyValidator\BodyValidator;
 use League\OpenAPIValidation\PSR7\Validators\CookiesValidator\CookiesValidator;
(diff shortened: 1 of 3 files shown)


src/PSR7/RequestValidator.php
```diff
@@ -102,4 +107,14 @@ private function findMatchingOperations(RequestInterface $request) : array
...
     }
+
+    /**
+     * Checks whether request matches any path in schema
+     */
+    private function hasMatchingPaths(RequestInterface $request) : bool
+    {
+        $pathFinder = new PathFinder($this->openApi, (string) $request->getUri(), $request->getMethod());
+
+        return $pathFinder->getPathMatches() !== [];
```
scaytrase, line 118: Actually I think we can make it as fast caching data in `PathFinder` (it's statefyl at the moment, so we can do that easily). For this we need to.  Use same `PathFinder` twice

jimdelois (approved):
LGTM - thanks!

PR #96, opened 2020-11-27
PHP 8 Support

### LINKED ISSUE #91: Support PHP 8
Hi,

PHP 8 will be released at the end of this month. It'd be cool if we can add support for it.

I began to investigate what needs to be done to support PHP 8. Most packages we depend on have releases that support PHP 8. Here are the non-compatible packages and issues I found:

- https://github.com/cebe/php-openapi currently does not have support. I created there https://github.com/cebe/php-openapi/issues/81 to track the progress.

- https://github.com/dflydev/dflydev-fig-cookies I created https://github.com/dflydev/dflydev-fig-cookies/issues/38 and https://github.com/dflydev/dflydev-fig-cookies/pull/39 . Waiting for response. But don't have much hope. Can we find an alternative?

- https://github.com/php-cache/array-adapter does not support PHP 8. I'm not sure they are open to new PRs. Looks like it didn't receive any maintenance over the last 2 years. Symfony's [solution](https://symfony.com/doc/current/components/cache.html) can be used instead.

That's all I guess. What do you think?

canvural:
This PR adds support for PHP 8 by removing `dflydev/fig-cookies` and replacing it with `hansott/psr7-cookies` and some custom code, and replaces `cache/array-adapter` with `symfony/cache`

Most of the changes are coming from style changes. So I made separate commits. [Here](https://github.com/thephpleague/openapi-psr7-validator/commit/a3bc3376953036e6f668149663eef9c44a10ad43) is the most relevant commit.

I think we don't need to wait for https://github.com/cebe/php-openapi/issues/81 to merge this. That library is already compatible with PHP8, only issues were in dev dependencies there.

Fixes #91

### DIFF
--- a/src/PSR7/Validators/CookiesValidator/RequestCookieValidator.php
+++ b/src/PSR7/Validators/CookiesValidator/RequestCookieValidator.php
@@ -5,8 +5,6 @@ declare(strict_types=1);
 namespace League\OpenAPIValidation\PSR7\Validators\CookiesValidator;
 
 use cebe\openapi\spec\Parameter;
-use Dflydev\FigCookies\Cookie;
-use Dflydev\FigCookies\Cookies;
 use League\OpenAPIValidation\PSR7\Exception\Validation\InvalidCookies;
 use League\OpenAPIValidation\PSR7\MessageValidator;
 use League\OpenAPIValidation\PSR7\OperationAddress;
@@ -17,6 +15,10 @@ use League\OpenAPIValidation\Schema\SchemaValidator;
 use Psr\Http\Message\MessageInterface;
 use Psr\Http\Message\RequestInterface;
 use Webmozart\Assert\Assert;
+
+use function array_key_exists;
+use function explode;
+use function implode;
 
 class RequestCookieValidator implements MessageValidator
 {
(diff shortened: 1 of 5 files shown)


src/Foundation/ArrayHelper.php
```diff
@@ -20,7 +20,7 @@ final class ArrayHelper
      *
      * @param mixed[] $arr
      */
-    public static function isAssoc(array $arr) : bool
```
scaytrase, line 23: why these have changed? despite the fact I personally don't like it, I see no relevant changes in phpcs config. Most of the changed CS contains only this kind of changes

canvural: `doctrine/coding-standard` was updated. And rules inside there changed to cause this. Do you want me to exclude that rule?

scaytrase:
Is it possible to move CS fixes to separate PR? I don't want to drop them at all, but I think we must keep PR payload simple to understand. CS is always a debatable thing and I don't want it to prevent this from merging

canvural:
Point of having agreed on a code style is to stick to it and forget it. So if we decided to use `doctrine/coding-standard` at the beginning IMO we should just stick with it and let it do its thing and not waste time debating on it.

Also, I made separate commits for code style changes and other stuff to make it easier to understand. You can see the most relevant commit in the PR description.

But anyway, I can remove the code style changes from this PR.

scaytrase:
Yea, I'm not against applying updated CS rules, I'm mostly agains having 180+ changed files with payload of 4-5 that really corresponds to PR title. So yes, it would be nice to have to separate PRs for that

canvural:
Oh, just realized the problem. `doctrine/coding-standard` 8.x is the only release that supports PHP 8. So first the code style PR needs to be merged, then this PR.

scaytrase:
There is no problem on that I think

canvural:
Ok, updated the PR. Currently, it'll fail cause it can't install `"doctrine/coding-standard": "^6.0"` but that will be fixed by #97

src/PSR7/Validators/CookiesValidator/RequestCookieValidator.php
```diff
@@ -59,22 +57,40 @@ private function checkRequiredCookies(Cookies $cookies, OperationAddress $addr)
...
         }
     }
+
+    /**
+     * @return string[]
+     */
+    private function getCookiesFromMessage(MessageInterface $message): array
+    {
+        // Needed in case it is an array of cookies.
+        $cookieString = implode('; ', $message->getHeader('Cookie'));
```
sunkan, line 84: Can't you use `$message->getHeaderLine()`?

canvural: That uses `implode` with `, ` We don't want that here. So it s the same as that one only glue is different.

scaytrase:
Can you please rebase this one too?

canvural:
@scaytrase Done

scaytrase:
Yay, it even passes nightly travis branch now!

scaytrase:
released as https://github.com/thephpleague/openapi-psr7-validator/releases/tag/0.14

canvural:
Thank you! @scaytrase

scaytrase:
Thank you @canvural for your contribution!

PR #97, opened 2020-11-27
Update `doctrine/coding-standard` to 8.x

canvural:
This PR updates `doctrine/coding-standard` to 8.x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
     },
     "require-dev": {
         "cache/array-adapter": "^1.0",
-        "doctrine/coding-standard": "^6.0",
+        "doctrine/coding-standard": "^8.0",
         "guzzlehttp/psr7": "^1.5",
         "phpunit/phpunit": "^7|^8"
     },
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -13,5 +13,11 @@
     <file>tests</file>
 
     <!-- Include full Doctrine Coding Standard -->
-    <rule ref="Doctrine"/>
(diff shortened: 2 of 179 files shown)


scaytrase (approved):
Point for discussion. Should we keep rules when update changes them or we just strictly follow upstream ruleset.
For example `) : typehint`. We can change local ruleset in order to keep codebase the same as we used to work with, or we just want have a consistent codebase without any general requirements on how it looks.

I'm personally OK with new codestyle, I like `): typehint` format.

sunkan:
I think it's easier to just follow the upstream change otherwise after a couple of update you have your very own code style :smile: 

Also I prefer the `): typehint` format. It's the same we use at work

scaytrase:
> you have your very own code style
And it's exact the same as at the start of the project, that's the point. I'm ok with ANY codestyle, even very own one. But between stable own one and changing foreign I'd rather choose stable own.

sunkan:
Yes I see your point.

I would say that we should probably only think about changing code style with major versions. 

But since we haven't reached version 1 yet. I'm still fine with this update

scaytrase:
I think codestyle has nothing to do with versioning since any codestyle change is non-breaking change. You really don't care which codestyle class you are extending or implementing )
Codestyle itself is important only for developing community, not for users

scaytrase:
@canvural can you please reapply CS fixes to current master? I'll merge this after

canvural:
@scaytrase Done :+1:

scaytrase:
Thanks!

Issue #98, opened 2020-11-30
Dictionary's are not currently supported although openapi v3 supports them


rlweb:
It seems OpenAPI v3 supports dictionaries as shown here: https://swagger.io/docs/specification/data-models/dictionaries/

Here is an example valid schema:
```
type: object
additionalProperties: false
properties:
  identifiers:
    description: Identifiers
    type: array
    items:
      type: object
      additionalProperties:
        type: string
        maxLength: 128
        nullable: true
      example:
        ean: "5039314069422"
        sku: SKU001
```

The above currently doesn't work due to the ArrayHelper::isAssoc call here: https://github.com/thephpleague/openapi-psr7-validator/blob/master/src/Schema/Keywords/Type.php#L49

I'll try and have a go at creating a PR for this! I'll let you know how I get on

Thanks for the great package btw!

scaytrase:
> Here is an example valid schema:

The problem is that your example is invalid schema. `type: string` cannot have `additionalProperties` set. Only `object` can have this (since only `object`) have `properties` at all.

https://swagger.io/docs/specification/data-models/data-types/#free-form

Even in the docs you linked there is no such example with string and additional properties. The schema you are trying to write is


```yaml
type: object
additionalProperties: false
properties:
  identifiers:
    description: Identifiers
    type: object
    additionalProperties:
        type: string
    example: 
      ean: "5039314069422" 
      sku: ajksdhakjdhsak
```

This exact example actually is at the end of page you linked https://swagger.io/docs/specification/data-models/dictionaries/


this will allow
```json
{
  "identifiers": {
    "ean": "5039314069422",
    "sku": "ajksdhakjdhsak",
    "other": "some other identifier"
  }
}

rlweb:
@scaytrase Ahhh, oops thanks for your help!

scaytrase:
No problem!

PR #99, opened 2020-11-30
Allow for dictionary/hashmaps by the spec [#98]

rlweb:


tests/Schema/Keywords/TypeTest.php
```diff
@@ -93,4 +93,24 @@ public function invalidDataProvider(): array
...
         ];
     }
+
+    public function testItValidatesDictionaryTypeGreen(): void
+    {
+        $spec = <<<SPEC
+schema:
+  type: array
+  items:
+    type: string
```
scaytrase, line 103: `string` type cannot have additional properties. only `object` type is allowed to do so. can you please provide an example test with samples from linked docs in order to check that the problem exists?

src/Schema/Keywords/Type.php
```diff
@@ -46,10 +65,29 @@ public function validate($data, string $type, ?string $format = null): void
 
                 break;
             case CebeType::ARRAY:
-                if (! is_array($data) || ArrayHelper::isAssoc($data)) {
+                $additionalProperties = $this->parentSchema->items
```
scaytrase, line 68: I'm not sure whether I got this expression. `items` could be a scalar type which cannot have additional properties

scaytrase (changes requested):
I think the intention is invalid, let's get back to original issue to discuss

Issue #102, opened 2020-12-10
Please support respect/validation 2.1+


franksl:
Hi,
Is it possible to support newer versions of respect/validation library?
I'm using ^2.1 version of respect/validation in my code and found a conflict when trying to require openapi-psr7-validator.
I'm aware that respect/validation requires php 7.3 while this requires php 7.2, maybe an upgrade of php version could be useful here too.
Thanks,
Frank

scaytrase:
It depends on how different package behaves on 1.x and 2.x. For some packages and some usages it's possible to support several versions of dependencies, for some it's not.

The perfect option is to have `respect/validation: ^1 || ^2` in order to get maximum users.

Would you like to try upgrade `respect/validation` here in order to check, how much codebase changes it require? I think we may  take one or another way of sovling this based on that information.

franksl:
I created a pull request, respect/validation has some backwards incompatible changes, so either v1 or v2 can be supported. Hope this helps,
Frank

PR #103, opened 2020-12-11
Support for respect/validation v2.1 (php >=7.3)

franksl:
Hi,
As discussed in issue #102 I did some changes to support respect/validation v2.1.
This will require minimum php version 7.3, which I think is reasonalbe nowadays.
respect/validation 2 has a couple of backward incompatible changes so I had to change all the references to the ExceptionInterface interface with new ones.
I noticed that the cases where respect/validation may raise an exception are not fully tested in the test suite, so I send this pul request as is. The actual test suite has success on all tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         }
     },
     "require": {
-        "php": ">=7.2",
+        "php": ">=7.3",
         "ext-json": "*",
         "cebe/php-openapi": "^1.3",
         "hansott/psr7-cookies": "^3.0.2",
@@ -29,7 +29,7 @@
         "psr/cache": "^1.0",
         "psr/http-message": "^1.0",
         "psr/http-server-middleware": "^1.0",
-        "respect/validation": "^1.1.3",
+        "respect/validation": "^2.1",
         "riverline/multipart-parser": "^2.0.3",
         "webmozart/assert": "^1.4"
     },
--- a/src/Schema/Keywords/MultipleOf.php
+++ b/src/Schema/Keywords/MultipleOf.php
@@ -6,7 +6,6 @@ namespace League\OpenAPIValidation\Schema\Keywords;
 
 use League\OpenAPIValidation\Schema\Exception\InvalidSchema;
 use League\OpenAPIValidation\Schema\Exception\KeywordMismatch;
-use Respect\Validation\Exceptions\ExceptionInterface;
 use Respect\Validation\Validator;
 
 use function sprintf;
@@ -25,9 +24,9 @@ class MultipleOf extends BaseKeyword
     public function validate($data, $multipleOf): void
     {
         try {
-            Validator::numeric()->assert($data);
-            Validator::numeric()->positive()->assert($multipleOf);
-        } catch (ExceptionInterface $e) {
+            Validator::numericVal()->assert($data);
+            Validator::numericVal()->positive()->assert($multipleOf);
+        } catch (\Respect\Validation\Exceptions\Exception $e) {
             throw InvalidSchema::becauseDefensiveSchemaValidationFailed($e);
         }
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ docs/
 .phpunit*
 composer.lock
 .phpcs*
+.vscode
--- a/src/PSR7/Validators/SerializedParameter.php
+++ b/src/PSR7/Validators/SerializedParameter.php
@@ -12,7 +12,6 @@ use League\OpenAPIValidation\Schema\Exception\ContentTypeMismatch;
 use League\OpenAPIValidation\Schema\Exception\InvalidSchema;
 use League\OpenAPIValidation\Schema\Exception\SchemaMismatch;
 use League\OpenAPIValidation\Schema\Exception\TypeMismatch;
-use Respect\Validation\Exceptions\ExceptionInterface;
 use Respect\Validation\Validator;
 
 use function explode;
@@ -72,7 +71,7 @@ final class SerializedParameter
             }
 
             Validator::length(1, 1)->assert($content);
-        } catch (ExceptionInterface $e) {
+        } catch (\Respect\Validation\Exceptions\Exception $e) {
             // If there is a `schema`, `content` must be empty.
             // If there isn't a `schema`, a `content` with exactly 1 property must exist.
             // @see https://swagger.io/docs/specification/describing-parameters/#schema-vs-content
--- a/src/Schema/Keywords/AllOf.php
+++ b/src/Schema/Keywords/AllOf.php
@@ -9,7 +9,6 @@ use League\OpenAPIValidation\Schema\BreadCrumb;
 use League\OpenAPIValidation\Schema\Exception\InvalidSchema;
 use League\OpenAPIValidation\Schema\Exception\SchemaMismatch;
 use League\OpenAPIValidation\Schema\SchemaValidator;
-use Respect\Validation\Exceptions\ExceptionInterface;
 use Respect\Validation\Validator;
 
 final class AllOf extends BaseKeyword
@@ -46,7 +45,7 @@ final class AllOf extends BaseKeyword
         try {
             Validator::arrayVal()->assert($allOf);
             Validator::each(Validator::instance(CebeSchema::class))->assert($allOf);
-        } catch (ExceptionInterface $exception) {
+        } catch (\Respect\Validation\Exceptions\Exception $exception) {
             throw InvalidSchema::becauseDefensiveSchemaValidationFailed($exception);
         }
 
--- a/src/Schema/Keywords/AnyOf.php
+++ b/src/Schema/Keywords/AnyOf.php
@@ -11,7 +11,6 @@ use League\OpenAPIValidation\Schema\Exception\KeywordMismatch;
 use League\OpenAPIValidation\Schema\Exception\NotEnoughValidSchemas;
 use League\OpenAPIValidation\Schema\Exception\SchemaMismatch;
 use League\OpenAPIValidation\Schema\SchemaValidator;
-use Respect\Validation\Exceptions\ExceptionInterface;
 use Respect\Validation\Validator;
 
 class AnyOf extends BaseKeyword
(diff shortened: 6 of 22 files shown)


composer.json
```diff
@@ -21,15 +21,15 @@
         }
     },
     "require": {
-        "php": ">=7.2",
+        "php": ">=7.3",
```
scaytrase, line 24: this change is not really required. yes, it's implied by the dependency, but the project itself does not require 7.3

src/Schema/Keywords/MultipleOf.php
```diff
@@ -25,9 +24,9 @@ class MultipleOf extends BaseKeyword
     public function validate($data, $multipleOf): void
     {
         try {
-            Validator::numeric()->assert($data);
-            Validator::numeric()->positive()->assert($multipleOf);
-        } catch (ExceptionInterface $e) {
+            Validator::numericVal()->assert($data);
```
scaytrase, line 27: I don't see real changes other than `ExceptionInterace` -> `Exception` and `numeric` -> `numericVal`. I think this can be tried to overcome with some kind of compatibility hacks to support both versions

franksl:
Hi,
Please take my pull request as a simple analisys, I had to integrate the libraries in an important project so I did a quick investigation and fix to make it work.
Surely there are ways to make both v1 and v2 work, I had to make it work quickly so I went the dirt way :)

composer.json
franksl: Onestly I don't know why and if respect/validation requires php 7.3 instead of 7.2, but its composer.json has this:
```
"php": "^7.3 || ^8.0",
```

src/Schema/Keywords/MultipleOf.php
sunkan: We shouldn't even need a hack this should work:
```
} catch (ExceptionInterface|\Respect\Validation\Exceptions\Exception $e) {
```

php doesn't care if the class your matching exists or not

composer.json
scaytrase: I just mean that changes in dependencies are not obligatory to make changes in project requirements. composer will resolve all minimul requirements by itself, so we can keep 7.2 here (since we may once change validator lib to another with another dependencies). 7.2 here means that the code of this library itself requires 7.2 to work, not for code of the dependencies

src/Schema/Keywords/MultipleOf.php
scaytrase: @franksl can you please try to update catch blocks instead of replace it? Also with imports instead of FQCN use? I  will think about `numeric` meanwhile

scaytrase: actually one way to solve it is to ask `numericVal` into 1.x version as an alias to `numeric`

franksl: > php doesn't care if the class your matching exists or not

I strongly disagree, sorry but if php lets you write names of classes that don't exist to me is a BIG flaw. I absolutely don't mean to be rude but taking advantage of what I consider flaws is not the way I would fix that.

composer.json
franksl: You're right, it seems that composer lets the installation go ahead iven with the 7.2 constraint.
This looks strange to me, but maybe I misunderstood something, I wonder what would happen if you call methods in respect/validation that require php 7.3...

src/Schema/Keywords/MultipleOf.php
sunkan: Do you have any other way to support both version then?

My fix is simple and it's supported by php so I don't see a problem with that.

franksl: If all the methods of respect/validation used by this library work on php 7.2 I would support only validation v2.
It's even simpler. 
Personal opinion: the fact that php supports what I consider bad practices does not mean that they should be encouraged. Surely I will not put my name on that. The easiest way is not always the right way.

sunkan: But that's not the problem. 

Lets say I'm using respect/validation v1 and this library. 
Your way of doing this breaks this for me because I can't upgrade because of blocking requirement on v2. 

If we support both 1 and 2 it would install v1 for me and v2 for you. 

And the library would continue to work for php7.2

scaytrase: > I would support only validation v2. It's even simpler.

This means than next day after rease we got a bug report for older release and we say "oh, dude, you need to update your php, half of your project and a lot more in order to get this bugfix because now we support only respect v2".

Other way is to support two versions simulataniously, which is a lot overhead, since then we should do this for each library which get a major update.

So the easiest way is to make any assumptions of your dependencies and make different code work for different parts of them.

Imaging having these code in C
```c
#ifdef REPSECT_VERSION_2
  do_some_new_stuff(args);
#else
 do_some_old_stuff(args);
#endif
```

Having class loader lazy is the only way to have optional dependencies. Otherwize any framework will carry all the dependencies it can support or we will get a hook hell for configuring plugin system (imagine any service would be a drupal with hundred of plugins)

franksl: Correct, but considering that the last release of respect/validation is 1 year and a half old I must assume that the v1 branch is highly abandoned in favor of the new one.
If the purpose of this library is to support all the old versions of all its dependencies well... good luck... this is not my home, I don't make any decisions, I'm just suggesting an upgrade that, to me, is reasonable.

scaytrase: If we are talking about end-user project (like complete system with business logic and so on) I'm totally agree with you. Doing weird stuff with classloader is ugly and is not welcome. But main purpose of the library is to provider smooth DX to the programmers who use it. And the never know whether we use classloader trick or not, because it's incapsualted in our private logic.

composer.json
scaytrase: Nothing. If this library require 7.2, respect/validation require 7.3 and you successfully installed this library - you using 7.3. Other cases are disallowed by composer. In that cases you will get only that versions of this libarry which have previous verions of respect/validation (with 7.2 for example).

src/Schema/Keywords/MultipleOf.php
franksl: If you agree to the fact that this is ugly and not welcome I would not place all this uglyness across the code base that also compromises clarity and testability.
I suggest a wrapper class for validation, this way you'll contain the "damage".

scaytrase: No, it's not not welcome for libs. I was talkin only about end-projects. How does the wrapper class helps us with testability?

franksl: You handle and test Validation exception only in the wrapper class e re-throw an internal exception (declared in THIS library) and catch the internal exception across the code. This way you need to check eventual class renames only in the wrapper and also can swap validation libraries if you want. 
The origin of the problem comes from the fact that when I installed v2 all the tests here were passing, but the library had some weird behaviours because the class in the catch sentences did not exist anymore.

franksl: Also it looks like that there is a phpversion() function (did not know that), that could be useful to make take some decisions. Composer is php code so if they can check version, so we do. I did not check if there are any problems with thread safety, etc, with these function, but I don't see why

scaytrase: see also https://www.php.net/manual/ru/reserved.constants.php
https://www.php.net/manual/ru/function.version-compare.php
very useful functions. there are also some mechanics like `class_exists`, and some utilities like https://github.com/Ocramius/PackageVersions (which was superseded by composer 2 IIUC)

franksl: Thanks for the links, I was aware of class_exists and all other versioning functions, but never investigated on retrieving the php exe version, good to know.

franksl: Hi,
Here's am example  for a validation wrapper, it's very rough and not tested, just to give an idea. DataValidation is a new exception class that should be created.
```
namespace League\OpenAPIValidation\Foundation;

use League\OpenAPIValidation\Foundation\Exception\DataValidation;
use Respect\Validation\Validator;

class DataValidator
{
    public function __callStatic($name, $arguments)
    {
        $v = self::getValidationMajorVersion();
        try {
            $n = $name;
            if ($name == 'numeric') {
                $v = self::getValidationMajorVersion();
                if ($v >= 2) {
                    $n = 'numericVal';
                }
            }
            call_user_func_array([Validator, $n], $arguments);
        } catch (ExceptionInterface|\Respect\Validation\Exceptions\Exception $ex) {
            throw new DataValidation($ex->getMessage(), $ex->getCode(), $ex);
        }
    }

    public static function getValidationMajorVersion()
    {
        if (version_compare(phpversion(), '7.3', '>=')) {
            return 2;
        } else {
            return 1;
        }
    }
}
```

scaytrase: Yea, looks good. But actually I'd rather prefer to do check via `method_exists` since it's less error prone (imagine respect/validation for some reason fix their 1.x branch and add support of php 7.3 - the code will break).

But it still has nothing to do with testability. But yea - it's all the same trick we want to apply, just beatyfied and coupled. Good work.

Some notes though:
* `DataValidation` isn't always correct exception here, since we also utilize validation as a defensive validations for schema state (which this lib is not responsible for) and arguments (which should be validated somewhere else and thus failure here is logic exception). So I'd rather suggest name it like `InvalidValue` or so to avoid consfused of what we actually validating

sunkan: This code would fail in the most awesome why if I'm running respect/validation 1 on php 7.3 or above

franksl: You're right, as I said it's just an example, I wrote it in 2 minutes just to express the idea

scaytrase: Yea, I forgot that you can just run validation:1 on php7.3 and up too (our travis\GA prooves it). So it won't work like this

scaytrase: > You're right, as I said it's just an example, I wrote it in 2 minutes just to express the idea

Sure, idea itself is good!

franksl: If you agree with this way I can take some time to work on a new PR as soon as possible and using method_exists, etc, otherwise I'll stay on issue #102 kindly asking to support validation v2 in some way

PR #106, opened 2020-12-12
Add static analysis with PHPStan

canvural:
This PR adds static analysis with PHPStan.

Code is already very well written, so there were no big issues. And this PR will also help with making sure it stays this way.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,8 +4,8 @@ on: [push, pull_request]
 
 jobs:
 
-  build:
-    name: Build
+  tests:
+    name: Tests
     runs-on: ubuntu-latest
     continue-on-error: ${{ matrix.experimental }}
     strategy:
@@ -68,3 +68,27 @@ jobs:
 
       - name: Run tests
         run: ./vendor/bin/phpcs
+
+  static-analysis:
+    name: Static analysis
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php:
+          - '7.2'
+
+    steps:
+      - name: Set up PHP
+        uses: shivammathur/setup-php@2.1.0
+        with:
+          php-version: ${{ matrix.php }}
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Download dependencies
+        run: composer update --no-interaction --prefer-dist --optimize-autoloader --prefer-stable
+
+      - name: Run PHPStan
+        run: ./vendor/bin/phpstan analyse --memory-limit 512M
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,10 @@
     "require-dev": {
         "doctrine/coding-standard": "^8.0",
         "guzzlehttp/psr7": "^1.5",
+        "phpstan/extension-installer": "^1.0",
+        "phpstan/phpstan": "^0.12.59",
+        "phpstan/phpstan-phpunit": "^0.12.16",
+        "phpstan/phpstan-webmozart-assert": "^0.12.7",
         "phpunit/phpunit": "^7|^8|^9",
         "symfony/cache": "^5.1"
     },
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,77 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Parameter \\#4 \\$prev of static method League\\\\OpenAPIValidation\\\\PSR7\\\\Exception\\\\Validation\\\\InvalidCookies\\:\\:becauseValueDoesNotMatchSchema\\(\\) expects League\\\\OpenAPIValidation\\\\Schema\\\\Exception\\\\SchemaMismatch, Throwable\\|null given\\.$#"
+			count: 1
+			path: src/PSR7/Validators/CookiesValidator/ServerRequestCookieValidator.php
+
+		-
+			message: "#^Parameter \\#1 \\$openApiSpec of class League\\\\OpenAPIValidation\\\\PSR7\\\\PathFinder constructor expects cebe\\\\openapi\\\\spec\\\\OpenApi, cebe\\\\openapi\\\\SpecObjectInterface given\\.$#"
+			count: 2
+			path: tests/FromCommunity/Issue79Test.php
+
+		-
+			message: "#^Generator expects value type string, array\\<int, string\\> given\\.$#"
+			count: 7
+			path: tests/FromCommunity/Issue79Test.php
+
+		-
+			message: "#^Parameter \\#1 \\$openApiSpec of class League\\\\OpenAPIValidation\\\\PSR7\\\\PathFinder constructor expects cebe\\\\openapi\\\\spec\\\\OpenApi, cebe\\\\openapi\\\\SpecObjectInterface given\\.$#"
+			count: 4
+			path: tests/PSR7/PathFinderTest.php
+
+		-
+			message: "#^Method League\\\\OpenAPIValidation\\\\Tests\\\\PSR7\\\\Validators\\\\BodyValidator\\\\MultipartValidatorTest\\:\\:dataProviderMultipartRed\\(\\) should return array\\<array\\<string, string\\>\\> but returns array\\<int, array\\<int, string\\>\\>\\.$#"
+			count: 1
+			path: tests/PSR7/Validators/BodyValidator/MultipartValidatorTest.php
+
+		-
+			message: "#^Method League\\\\OpenAPIValidation\\\\Tests\\\\PSR7\\\\Validators\\\\BodyValidatorTest\\:\\:dataProviderGreen\\(\\) should return array\\<array\\<string, string\\>\\> but returns array\\<int, array\\<int, League\\\\OpenAPIValidation\\\\PSR7\\\\OperationAddress\\|string\\>\\>\\.$#"
+			count: 1
+			path: tests/PSR7/Validators/BodyValidatorTest.php
+
+		-
+			message: "#^Method League\\\\OpenAPIValidation\\\\Tests\\\\PSR7\\\\Validators\\\\BodyValidatorTest\\:\\:dataProviderRed\\(\\) should return array\\<array\\<string, string\\>\\> but returns array\\<int, array\\<int, string\\>\\>\\.$#"
+			count: 1
+			path: tests/PSR7/Validators/BodyValidatorTest.php
+
+		-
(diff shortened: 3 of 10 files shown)


src/PSR7/SpecFinder.php
```diff
@@ -231,6 +231,8 @@ public function findResponseSpec($addr): ResponseSpec
      * @return Header[]
      *
      * @throws NoPath
+     *
+     * @phpstan-return array<string, Header|Parameter>
```
scaytrase, line 235: AFAIK modern IDEs understand plain `@return` with the same contents as well, no need in special phpstan prefix. Can you check it ?

src/PSR7/SpecFinder.php
```diff
@@ -244,7 +246,9 @@ public function findHeaderSpecs(OperationAddress $addr): array
         // 1. Collect operation level headers from "parameters" keyword
         // An API call may require that custom headers be sent with an HTTP request. OpenAPI lets you define custom
         // request headers as in: header parameters.
+        /** @var array<string, Header> $headerSpecs */
```
scaytrase, line 249: this is invalid according to later usage and return type. you cannot put `Parameter` into `array<string, Header>`. So it should be same as return type right here

src/PSR7/SpecFinder.php
```diff
@@ -262,6 +266,8 @@ public function findHeaderSpecs(OperationAddress $addr): array
         // 2. Collect path-level headers from "parameters" keyword
         // Path level params are fall-backs
         $pathSpec = $this->findPathSpec($addr);
+
+        /** @var Parameter $p */
```
scaytrase, line 270: This can't be inferred?

phpstan-baseline.neon
```diff
@@ -0,0 +1,77 @@
...
+			count: 7
+			path: tests/FromCommunity/Issue79Test.php
+
+		-
+			message: "#^Parameter \\#1 \\$openApiSpec of class League\\\\OpenAPIValidation\\\\PSR7\\\\PathFinder constructor expects cebe\\\\openapi\\\\spec\\\\OpenApi, cebe\\\\openapi\\\\SpecObjectInterface given\\.$#"
+			count: 4
+			path: tests/PSR7/PathFinderTest.php
+
+		-
+			message: "#^Method League\\\\OpenAPIValidation\\\\Tests\\\\PSR7\\\\Validators\\\\BodyValidator\\\\MultipartValidatorTest\\:\\:dataProviderMultipartRed\\(\\) should return array\\<array\\<string, string\\>\\> but returns array\\<int, array\\<int, string\\>\\>\\.$#"
```
scaytrase, line 24: i think this kind of problems can be fixed as well

composer.json
```diff
@@ -36,6 +36,10 @@
     "require-dev": {
         "doctrine/coding-standard": "^8.0",
         "guzzlehttp/psr7": "^1.5",
+        "phpstan/extension-installer": "^1.0",
+        "phpstan/phpstan": "^0.12.59",
+        "phpstan/phpstan-phpunit": "^0.12.16",
```
scaytrase, line 41: just note: would be nice to have `phphstan strict` here later

.github/workflows/ci.yml
```diff
@@ -68,3 +68,30 @@ jobs:
 
       - name: Run tests
         run: ./vendor/bin/phpcs
+
+  static-analysis:
+    name: Static analysis
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
```
scaytrase, line 77: is phpstan output really differs for all of these languages? I think we only can fix that one that refers to minimum php version we allow

.github/workflows/ci.yml
```diff
@@ -68,3 +68,30 @@ jobs:
...
+          php-version: ${{ matrix.php }}
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Download dependencies
+        run: composer update --no-interaction --prefer-dist --optimize-autoloader --prefer-stable
+
+      - name: Run PHPStan
+        run: ./vendor/bin/phpstan analyse --memory-limit 512M
```
scaytrase, line 97: I suggest to add explicit phpstan level parameter here in order to always know what level we are on currently

phpstan-baseline.neon
```diff
@@ -0,0 +1,77 @@
+parameters:
+	ignoreErrors:
```
scaytrase, line 2: do we need baseline at all? most of this errors are easy fix (like dataproviders). Others will be ignored forever until we bring phpstan to some key dependencies

src/PSR7/Validators/SecurityValidator.php
```diff
@@ -165,6 +167,8 @@ private function validateApiKeySecurityScheme(
         RequestInterface $request,
         SecurityScheme $securityScheme
     ): void {
+        Assert::isInstanceOf($request, ServerRequestInterface::class);
```
scaytrase, line 170: actually this should be guarded here. so all other signatures can just be fixed to be `ServerRequestInterface` directly, not via asserts

https://github.com/thephpleague/openapi-psr7-validator/pull/106/files#diff-be02d5fa32966a6c559f82d49d961f2236badc1858ca4f7c6ebc09b58387dbabR50-R52

scaytrase (changes requested):
Nice work! Need some polish and ready to go. Static analisys is always welcome

src/PSR7/Validators/SecurityValidator.php
canvural: Are you suggesting to change the signature of `validateApiKeySecurityScheme` method?

phpstan-baseline.neon
canvural: I think we do. The errors in baseline are not worth fixing I think. And it'll take a lot of time to fix them.

.github/workflows/ci.yml
canvural: Level is defined in `phpstan.neon` file

.github/workflows/ci.yml
canvural: Yeah, I think we can just run it on one version. Which version should we run on?

composer.json
canvural: Yeah, agree.

phpstan-baseline.neon
canvural: Maybe, but I think it doesn't worth the effort.

src/PSR7/SpecFinder.php
canvural: `$pathSpec->parameters` has type of `Parameter[]|Reference[]` So it PHPStan thinks it's `Parameter|Reference` So we need to force it to think it's just parameter.

src/PSR7/SpecFinder.php
canvural: Yeah, at least PHPStorm recognizes it correctly. Changed.

src/PSR7/SpecFinder.php
canvural: Fixed.

src/PSR7/Validators/SecurityValidator.php
scaytrase: exactly, it's private and guarded - why not?

.github/workflows/ci.yml
scaytrase: didn't noticed it. good!

.github/workflows/ci.yml
scaytrase: I prefer stick to lowest php version possible here in order not to accidently do checks for newer features

phpstan-baseline.neon
scaytrase: yea, maybe. at least from your side. maybe I'll fix them by myself later

src/PSR7/SpecFinder.php
scaytrase: does it work because we forced to resolve references beforehand? just to confirm it's not a potential bug (which is why phpstan actually comes here)

phpstan-baseline.neon
scaytrase: Yeah, sounds reasonable. I'll do it by myself later and drop baseline when I fix everything

src/PSR7/Validators/SecurityValidator.php
canvural: Well, it is guarded to be `RequestInterface`, not `ServerRequestInterface` [here.](https://github.com/thephpleague/openapi-psr7-validator/blob/master/src/PSR7/Validators/SecurityValidator.php#L48) So we would have to change it also there to `ServerRequestInterface` Do you want that?

src/PSR7/SpecFinder.php
canvural: I'm not sure. Maybe we don't have a test scenario that has headers with reference.

.github/workflows/ci.yml
canvural: Changed.

src/PSR7/Validators/SecurityValidator.php
scaytrase: I think it's better then adding more asserts which complicates reading

src/PSR7/SpecFinder.php
scaytrase: If we're not sure we can drop this one into baseline ignores too. I just don't want to hardcode things we are not sure in.

canvural: Ok, it's a kind of bug :smile: If you have a spec with header parameters with references and you use this the following code
```php
(new ValidatorBuilder())
    ->setSchemaFactory(new PrecreatedSchemaFactory(Reader::readFromYaml($yaml)))
    ->getServerRequestValidator()
    ->getSchema();
```

You get an error `Undefined property: cebe\openapi\spec\Reference::$in` at `src/PSR7/SpecFinder.php:250`

I would say this is an edge case. The majority of people are using JsonFactory or YamlFactory which resolves the references.

What do you think? Should I create a separate issue for this? Or we can ignore it?

scaytrase: It would be nice to file an issue **and** ignore this with phpstan until it's resolved. so yes, phpstan actually found a bug. that's I was talking about :D
👍

canvural: Ok. Reverted this. But funny PHPStan doesn't complain about it anymore. Weird..

src/PSR7/Validators/SecurityValidator.php
canvural: Changed.

scaytrase (approved):
👍 Great work! A lot can be improved later, but it's a good base to start from. Thanks a lot!

Issue #108, opened 2020-12-30
Let allow to use deepObject for arrays?


nektru:
I've needed to validate requests like `?id[]=1id[]=2` in my project but for now it is broken since validation becomes strict.

PR #109, opened 2020-12-30
Allow to use arrays with deepObject serialisation in the query string

nektru:
PR to issue #108

### DIFF
--- a/src/PSR7/Validators/SerializedParameter.php
+++ b/src/PSR7/Validators/SerializedParameter.php
@@ -17,6 +17,7 @@ use Respect\Validation\Validator;
 use function explode;
 use function in_array;
 use function is_array;
+use function is_bool;
 use function is_float;
 use function is_int;
 use function is_numeric;
@@ -158,8 +159,8 @@ final class SerializedParameter
     }
 
     /**
-     * @param mixed $value
-     * @param ?CebeSchema $schema - optional schema of value to convert it in case of DeepObject serialisation
+     * @param mixed           $value
+     * @param CebeSchema|null $schema - optional schema of value to convert it in case of DeepObject serialisation
      *
      * @return mixed
      */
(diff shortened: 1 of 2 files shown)


scaytrase:
Thanks @nektru, on a first look - good, but I need some more to inspect tests and logic. I'll do it a bit later

src/PSR7/Validators/SerializedParameter.php
```diff
@@ -192,4 +198,23 @@ public function getSchema(): CebeSchema
...
     }
+
+    protected function getChildSchema(CebeSchema $schema, string $key): ?CebeSchema
+    {
+        if ($schema->type === CebeType::OBJECT) {
+            if ($schema->properties[$key] ?? false) {
+                return $schema->properties[$key];
+            }
+
+            if (! is_bool($schema->additionalProperties) && $schema->additionalProperties instanceof CebeSchema) {
```
scaytrase, line 209: why `additionalProperties` equal to `true` is avoided here? This is also valid case with (in general) empty schema as a property spec (property can be anything)

nektru: Yes. It is a valid case, but we can't detect the concrete schema of a child type to convert the value to it. But also we no need to do any convertation there if we allow anything.

Alll three variants of the schema will be valid if we request smth like `path?id[a]=1&id[b]=2` :

```
id:
  type: object
  additionalProperties: true
```
```
id:
  type: object
  additionalProperties:
    type: integer
```
```
id:
  type: object
  additionalProperties:
    type: string
```

BTW For now I think I can just remove this check - it not unnecessary because object type checking is more important.

scaytrase: I think returning an `\stdClass` for `additionalProperties: true` as `id` value is purely valid here. Without any other guessing of property types

nektru: In case `additionalProperties: true` we should return null because no child schema is presented. So values will be casts without any type in `SerializedParameter::castToSchemaType($value, $type)` method . E.q. value will be used as-is. This situation is still valid because we haven't type in the schema to compare the type of value with.

scaytrase: Yea, null is processed as `any` indeed. This wasn't obvious from namings.

scaytrase (approved):

scaytrase:
Thanks @nektru ! Looks good! Let's give it a try 👍

Issue #110, opened 2021-01-03
Path parameter are not url decoded


porebskk:
Hello,  
first of all thanks for this fantastic library. I really appreciate all the work that you all are doing.  
  
I noticed a potential bug. I am using for one of my APIs a path parameter with the string format `e-mail`. Since the email contains the `@`-symbol, the value needs to be url encoded (`test%40example.com`). During the validation with this library, the `PathValidator` extracts the url encoded param, which will be later validated with a `TypeFormat`-validator. This will fail currently as the url encoded e-mail is not a valid e-mail format (missing the `@`-symbol).  
  
A possible fix would be to url decode the path params in `\League\OpenAPIValidation\PSR7\Validators\PathValidator::validateRequest` before they are passed to the `ArrayValidator`.

If the current behaviour is not intended, then I am happy to provide a pull request.

Best regards

scaytrase:
Hello, can we start with a PR with failing test case(s)?

PR #111, opened 2021-01-03
#110 test for url encoded path parameter

porebskk:
related to issue #110   
for now it only contains a failing test case

### DIFF
--- a/tests/PSR7/PathParametersTest.php
+++ b/tests/PSR7/PathParametersTest.php
@@ -98,4 +98,18 @@ final class PathParametersTest extends TestCase
         $validator->validate($request);
         $this->addToAssertionCount(1);
     }
+
+    public function testItValidatesBase64EncodedPathParameter(): void
+    {
+        $specFile = __DIR__ . '/../stubs/pathParams.yaml';
+
+        // evaluates to "/base64/c3ViamVjdHM%2FYQ%3D%3D"
+        $uri = '/base64/' . rawurlencode(base64_encode('subjects?a'));
+
+        $request = new ServerRequest('get', $uri);
+
+        $validator = (new ValidatorBuilder())->fromYamlFile($specFile)->getServerRequestValidator();
+        $validator->validate($request);
+        $this->addToAssertionCount(1);
+    }
 }
--- a/tests/PSR7/PathParametersTest.php
+++ b/tests/PSR7/PathParametersTest.php
@@ -9,6 +9,8 @@ use League\OpenAPIValidation\PSR7\Exception\Validation\InvalidPath;
 use League\OpenAPIValidation\PSR7\ValidatorBuilder;
 use PHPUnit\Framework\TestCase;
 
+use function urlencode;
+
 final class PathParametersTest extends TestCase
 {
     public function testItValidatesRequestQueryArgumentsGreen(): void
@@ -75,6 +77,22 @@ final class PathParametersTest extends TestCase
         // dot in path template must be handled with care
         $specFile = __DIR__ . '/../stubs/pathParams.yaml';
         $request  = new ServerRequest('get', '/number/99.json');
+
+        $validator = (new ValidatorBuilder())->fromYamlFile($specFile)->getServerRequestValidator();
+        $validator->validate($request);
+        $this->addToAssertionCount(1);
+    }
+
+    public function testItValidatesUrlEncodedPathParameter(): void
+    {
+        // Path parameter can be url encoded
+
+        $specFile = __DIR__ . '/../stubs/pathParams.yaml';
+
+        // results in "/email/openapi%40example.com"
+        $uri = '/email/' . urlencode('openapi@example.com');
+
+        $request = new ServerRequest('get', $uri);
 
         $validator = (new ValidatorBuilder())->fromYamlFile($specFile)->getServerRequestValidator();
         $validator->validate($request);
--- a/tests/stubs/pathParams.yaml
+++ b/tests/stubs/pathParams.yaml
@@ -79,3 +79,17 @@ paths:
       responses:
         204:
           description: No response
+  /email/{email}:
+    parameters:
+      - name: email
+        in: path
+        required: true
+        schema:
+          type: string
+          format: email
+    get:
+      summary: Read data
+      operationId: read-email-format
+      responses:
+        204:
+          description: No response
--- a/tests/stubs/pathParams.yaml
+++ b/tests/stubs/pathParams.yaml
@@ -93,3 +93,17 @@ paths:
       responses:
         204:
           description: No response
+  /base64/{base64}:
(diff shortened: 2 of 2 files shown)


tests/PSR7/PathParametersTest.php
```diff
@@ -80,4 +82,20 @@ public function testParsesFormat(): void
...
     }
+
+    public function testItValidatesUrlEncodedPathParameter(): void
+    {
+        // Path parameter can be url encoded
+
+        $specFile = __DIR__ . '/../stubs/pathParams.yaml';
+
+        // results in "/email/openapi%40example.com"
+        $uri = '/email/' . urlencode('openapi@example.com');
```
scaytrase, line 93: Can you verify, that `ServerRequest::createFromGlobals()` creates the request in the same way when processing generic requests? If I remove urlencode here - test passes, so it depens on how ServerRequest is being instantiated in a real world. Also we can refer to any kind of PSR specs in order to check whether it should operate encoded values or not

scaytrase: actually I did it by myself

I've created a simple script, which runs the following code and run it with built-in web server

```
<?php

use GuzzleHttp\Psr7\ServerRequest;

require_once __DIR__.'/vendor/autoload.php';

$request = ServerRequest::fromGlobals();
var_dump($request->getUri());
```

It' gives me the following output for url in browser `http://localhost/email/test@email.com`

```
lezhnev\openapi-psr7-validator\index.php:8:
object(GuzzleHttp\Psr7\Uri)[4]
  private 'scheme' => string 'http' (length=4)
  private 'userInfo' => string '' (length=0)
  private 'host' => string 'localhost' (length=9)
  private 'port' => null
  private 'path' => string '/email/test@email.com' (length=21)
  private 'query' => string '' (length=0)
  private 'fragment' => string '' (length=0)
```

scaytrase: If I manually create the request with URLencoding the same url - it gives me the following output
```
lezhnev\openapi-psr7-validator\index.php:11:
object(GuzzleHttp\Psr7\Uri)[30]
  private 'scheme' => string '' (length=0)
  private 'userInfo' => string '' (length=0)
  private 'host' => string '' (length=0)
  private 'port' => null
  private 'path' => string 'http%3A%2F%2Flocalhost%2Femail%2Ftest%40email.com' (length=49)
  private 'query' => string '' (length=0)
  private 'fragment' => string '' (length=0)
```

scaytrase: So I'd rather consider this as a misusing `ServerRequest` rather than a bug

porebskk: Yes, if the whole URI is url-encoded then you receive the second [output](https://github.com/thephpleague/openapi-psr7-validator/pull/111#discussion_r551867632). Take for example this URI-path `/email/example%40example.com`. I receive with your custom script following output:  
```
object(GuzzleHttp\Psr7\Uri)#9 (7) {
  ["scheme":"GuzzleHttp\Psr7\Uri":private]=>
  string(4) "http"
  ["userInfo":"GuzzleHttp\Psr7\Uri":private]=>
  string(0) ""
  ["host":"GuzzleHttp\Psr7\Uri":private]=>
  string(9) "localhost"
  ["port":"GuzzleHttp\Psr7\Uri":private]=>
  int(8080)
  ["path":"GuzzleHttp\Psr7\Uri":private]=>
  string(28) "/email/example%40example.com"
  ["query":"GuzzleHttp\Psr7\Uri":private]=>
  string(0) ""
  ["fragment":"GuzzleHttp\Psr7\Uri":private]=>
  string(0) ""
}
```
I see that the browser as well as the server can handle the @-symbol wihtout any problems. But what if someone want to provide a file path (e.g. /root/test.txt) as path parameter. In this case it is required to encode the path parameter or else it will be interpreted as URI-path.

porebskk: Another simpler example would be base64 encoded values in path parameter. Base64 consists of these allowed characters: A–Z, a–z, 0–9, +, / and =. The slash could lead to problems with routers or even the openapi validation as the slash would indicate a different path. So in order to mitigate this problem, one would url-encode the base64 encoded value. I pushed another test case for this scenario.

scaytrase: my point is that real-life  usages do not involve  manual urlencode to the uri parts. I think this should be reported first to the used PSR-7 implementation, since you've mentioned

> The PSR-7 specification also mentions as an example the encoding of "/" when they are not intended for path segments delimiter:

I've checked guzzle sources - it's not done there, so maybe it's a bug there.

porebskk: Okay i understand. Thanks for your time and help.

porebskk: Just as a last attempt, the specification of the PSR7 "getPath" states:  
```
     * The value returned MUST be percent-encoded, but MUST NOT double-encode
     * any characters.
```

So this indicates that the user needs to handle encoded output of "getPath".  Further, and this is what you are describing aswell, is the specification of "withPath":
```
     * Users can provide both encoded and decoded path characters.
     * Implementations ensure the correct encoding as outlined in getPath().
```

This refers to the input, not the output of the path. And with the PSR7 interface only allowing to provide one string as path, it is not possible for the library to detect, if for example a slash is meant as path segment delimiter or as user-value.  
So in regards of the limitation the specification provides, the manual urlencode is a required user action, if he wants to ensure that all his user-values are interpreted as such.

scaytrase: > Just as a last attempt

Don't get me wrong, I'm not trying to blindly decline your intends. I'm just trying to keep tests (and codebase) aligned with most popular usecases.

Can you try to add urldecode to the place you've mentioned in the original issue?

> Users can provide both encoded and decoded path characters.

For me the root cause problem is here, since this statment disables us to understand whether the string is urlencoded or not. For example we have string `@%40value`. Is it encoded `@@value` or user intended to write `@%40value` as it is?

sunkan: Hi 

I've been looking into this and where do you get that `@` is not a valid part of the path.
If you enter a url in a browser where the path contains `@` it will not be url encoded

And how I read the specification is that `@` is reserved in some of the components that make up the uri but not all. And all psr-7 implementation have implemented there path parsing accordingly and don't encode `@`

As can be seen in [guzzle/psr7](https://github.com/guzzle/psr7/blob/master/src/Uri.php#L642) where they actively ignore encoding `@` in the path

And as has been pointed out because this would be really hard to solve in a way that don't break stuff. 
If you know that the incoming request contains a url encoded path why don't you just add a middleware that decodes the path before pass it on to `openapi-psr7-validator`

PR #112, opened 2021-01-11
Support Respect/Validation v2

### LINKED ISSUE #102: Please support respect/validation 2.1+

t0mmy742:
Fixes #102 
Closes #103 (and based on these discussions)

Also completed `ignoreErrors` to phpstan.neon file since static analysis check is done with PHP 7.2 on GitHub CI.

This is the best thing we can do for the moment.
Adding a `DataValidator` [(see discussion)](https://github.com/thephpleague/openapi-psr7-validator/pull/103#discussion_r542231211) does not help since we loose a lot of things (auto-completion, static checks,...).
Possibly not the cleanest, but it supports v1 AND v2.
Plus we need to check with `class_exists` and not `method_exists`. These methods (numericVal,...) do not really exist, but use magic functions (__call, __callStatic).

One day, we could possibly think about dropping Respect/Validation v1 support (last release more than one year ago, 2019-05-28). Respect/Validation v2.0 supports PHP^7.2, while v2.1 supports PHP^7.3.

### DIFF
--- a/src/PSR7/Validators/SerializedParameter.php
+++ b/src/PSR7/Validators/SerializedParameter.php
@@ -11,6 +11,7 @@ use League\OpenAPIValidation\Schema\Exception\ContentTypeMismatch;
 use League\OpenAPIValidation\Schema\Exception\InvalidSchema;
 use League\OpenAPIValidation\Schema\Exception\SchemaMismatch;
 use League\OpenAPIValidation\Schema\Exception\TypeMismatch;
+use Respect\Validation\Exceptions\Exception;
 use Respect\Validation\Exceptions\ExceptionInterface;
 use Respect\Validation\Validator;
 
@@ -71,7 +72,7 @@ final class SerializedParameter
             }
 
             Validator::length(1, 1)->assert($content);
-        } catch (ExceptionInterface $e) {
+        } catch (Exception | ExceptionInterface $e) {
             // If there is a `schema`, `content` must be empty.
             // If there isn't a `schema`, a `content` with exactly 1 property must exist.
             // @see https://swagger.io/docs/specification/describing-parameters/#schema-vs-content
--- a/src/Schema/Keywords/Maximum.php
+++ b/src/Schema/Keywords/Maximum.php
@@ -6,9 +6,12 @@ namespace League\OpenAPIValidation\Schema\Keywords;
 
 use League\OpenAPIValidation\Schema\Exception\InvalidSchema;
 use League\OpenAPIValidation\Schema\Exception\KeywordMismatch;
+use Respect\Validation\Exceptions\Exception;
 use Respect\Validation\Exceptions\ExceptionInterface;
+use Respect\Validation\Rules\NumericVal;
 use Respect\Validation\Validator;
 
+use function class_exists;
 use function sprintf;
 
 class Maximum extends BaseKeyword
@@ -39,9 +42,14 @@ class Maximum extends BaseKeyword
     public function validate($data, $maximum, bool $exclusiveMaximum = false): void
     {
         try {
-            Validator::numeric()->assert($data);
-            Validator::numeric()->assert($maximum);
-        } catch (ExceptionInterface $e) {
+            if (class_exists(NumericVal::class)) {
+                Validator::numericVal()->assert($data);
+                Validator::numericVal()->positive()->assert($maximum);
+            } else {
+                Validator::numeric()->assert($data);
+                Validator::numeric()->positive()->assert($maximum);
+            }
+        } catch (Exception | ExceptionInterface $e) {
             throw InvalidSchema::becauseDefensiveSchemaValidationFailed($e);
         }
 
--- a/src/Schema/Keywords/Minimum.php
+++ b/src/Schema/Keywords/Minimum.php
@@ -6,9 +6,12 @@ namespace League\OpenAPIValidation\Schema\Keywords;
 
 use League\OpenAPIValidation\Schema\Exception\InvalidSchema;
 use League\OpenAPIValidation\Schema\Exception\KeywordMismatch;
+use Respect\Validation\Exceptions\Exception;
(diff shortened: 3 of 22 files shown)


src/PSR7/Validators/SerializedParameter.php
```diff
@@ -71,7 +72,7 @@ public static function fromSpec(CebeParameter $parameter): self
             }
 
             Validator::length(1, 1)->assert($content);
-        } catch (ExceptionInterface $e) {
+        } catch (Exception | ExceptionInterface $e) {
```
elazar, line 75: [Union types](https://wiki.php.net/rfc/union_types_v2) were added in PHP 8, so I think this change would make this class and any others with a similar change in this PR unusable in older versions of PHP.

canvural: This is not union types. Since PHP 7.1 you can specify multiple exceptions with pipe separator. https://www.php.net/manual/en/language.exceptions.php

elazar: Ah! I stand corrected, and sorry for the noise. I learned something today and appreciate the explanation. :smile:

src/Schema/Keywords/Maximum.php
```diff
@@ -39,9 +42,14 @@ class Maximum extends BaseKeyword
     public function validate($data, $maximum, bool $exclusiveMaximum = false): void
     {
         try {
-            Validator::numeric()->assert($data);
-            Validator::numeric()->assert($maximum);
-        } catch (ExceptionInterface $e) {
+            if (class_exists(NumericVal::class)) {
```
scaytrase, line 45: Sorry for long delay. @t0mmy742 can you please try to implement a wrapper discussed here (but with a `method_exists` check instead of version check) ?

Or at least replace class check to the `method_exists` here (and other similar places) since we are not relying on checked class here but on the method in the `Validator` class

scaytrase (changes requested):

src/Schema/Keywords/Maximum.php
scaytrase: Oh, I've read you note on PR body. Yeah, maybe implementing `DataValidator` is not worth here, but at least we should use `method_exists` instead of `class_exists` here.

t0mmy742: `method_exists` can not find this method since it is called with magic functions (`__call` and `__callStatic`). Using `is_callable` function should do the work, if you're ok with this 😅

scaytrase: uff. yea,  I just don't want to rely on the check which result is not being used. I.e they can delete or rename this class next version but keep `numericVal` static method with some kind of workaround. check will fail but code actually keep working.
let's try `is_callable`. If it's not working - well we can keep this as is

t0mmy742: I can't modify right now, I will try after work, or next days

t0mmy742: @scaytrase I just tested with `is_callable`. It always returns true. In fact, we can call a method with every name, since it uses `__callStatic` function. 
About them renaming or deleting this class, it should not happen since it's part of their public API, and that would mean they would be creating a BC break (for minor version at least).
If you really don't like this method, you should know that calling `Validator::method()` (with `method` being an incorrect name) result in throwing a `\Respect\Validation\Exceptions\ComponentException` exception (v1 AND v2). We could do something like this:
```php
try {
    try {
        Validator::numericVal()->assert($data);
        Validator::numericVal()->positive()->assert($maximum);
    } catch (ComponentException$e) {
        Validator::numeric()->assert($data);
        Validator::numeric()->positive()->assert($maximum);
    }
} catch (Exception | ExceptionInterface $e) {
    throw InvalidSchema::becauseDefensiveSchemaValidationFailed($e);
}
```

It's up to you to decide what you prefer :)

scaytrase: No the last one is perfomance-heavy, since you need to call a method, raise an exception and process it every time you have v1 installed. let stick to original plan then

scaytrase (approved):
LGTM 👍

scaytrase:
Thanks @t0mmy742 !

src/Schema/Keywords/Minimum.php
```diff
@@ -39,9 +42,14 @@ class Minimum extends BaseKeyword
     public function validate($data, $minimum, bool $exclusiveMinimum = false): void
     {
         try {
-            Validator::numeric()->assert($data);
-            Validator::numeric()->assert($minimum);
-        } catch (ExceptionInterface $e) {
+            if (class_exists(NumericVal::class)) {
+                Validator::numericVal()->assert($data);
+                Validator::numericVal()->positive()->assert($minimum);
```
samuelnogueira, line 47: This change added a new validation which doesn't allow **minimum** values to be equal, or lower than zero (see https://respect-validation.readthedocs.io/en/latest/rules/Positive/).
openapi specification seems to allow zero or negative values for both `minimum` and `maximum`, although not explicitly.
Was this intentional?

scaytrase: I think it's a copypaste from `MultipleOf`. Would you mind to submit a PR? I think there are some other problems of the same kind

scaytrase: Also maybe we need more tests around this, since this had pass unseen

samuelnogueira: Sure! I spotted only **minimum** and **maximum** with this extra rule, I'll add any others if I find them.

PR #114, opened 2021-01-13
Fixes #113 Update outdated README caching example

### LINKED ISSUE #113: Outdated caching example?
In the [Caching Layer / PSR-6 Support section](https://github.com/thephpleague/openapi-psr7-validator#caching-layer--psr-6-support) of the README, the bottom of the code example contains this code.

```php
# or
\League\OpenAPIValidation\PSR15\ValidationMiddleware::fromYamlFile($yamlFile, $cachePool);
```

The [`ValidationMiddleware` class](https://github.com/thephpleague/openapi-psr7-validator/blob/master/src/PSR15/ValidationMiddleware.php) doesn't have a parent class or use traits and does not contain a declaration of a static `fromYamlFile()` method. It appears that it did previously in 4d67276c8da6d1811544ec4400152ee37cdf6817, but this was removed in 7de08020dde0091686790031328b7c2db020dfc0.

The only class that does contain a method with the name `fromYamlFile()` is `ValidatorBuilder`, but that method implementation is not static.

Is the code referenced above simply an outdated example that wasn't removed as part of 7de08020dde0091686790031328b7c2db020dfc0, and should be removed? Or should it be modified somehow?

elazar:


scaytrase (approved):

scaytrase:
Thanks @elazar !

PR #117, opened 2021-01-20
`multipleOf` is not validating correctly for value `0.01`

garethellis36:
Good afternoon :)

Thanks for your continued work on this useful library.

I have come across a small bug in my work today. The validator incorrectly throws `KeywordMismatch` within `MultipleOf::validate()` when given a data value of `283.66` and a multipleOf value of `0.01`. 0.01 *is* a multiple of 283.66, and according to JSON Schema, `multipleOf` can be any positive number.

At the moment, this PR is just a failing test, I haven't added a fix yet. I need to get my head around the fun of floating point maths in PHP :) But in the meantime, if you have any suggestions on how to address this, I'm happy to implement.

### DIFF
--- a/tests/Schema/Keywords/MultipleOfTest.php
+++ b/tests/Schema/Keywords/MultipleOfTest.php
@@ -20,6 +20,9 @@ final class MultipleOfTest extends SchemaValidatorTest
             [10.0, 2],
             [10, .5],
             [9.9, .3],
+            [283, 0.01],
+            [283.6, 0.01],
+            [283.66, 0.01],
         ];
     }
 


garethellis36:
This 3v4l snippet demonstrates why it is failing: https://3v4l.org/ASg6k

sunkan:
I don't know if we should fix this because the open api specification says

> `multipleOf` may be used with floating-point numbers, **but in practice this can be unreliable due to the limited precision or floating point math.**

And since this a problem with the language, the only work around would be to round the number to some specific decimal point and that might cause more problems then it worth

scaytrase:
That's true. But we still can think whether we can improve without being less stable. I think `multipleOf` is valid example for money application (despite of the fact I always suggest use integers for money)

garethellis36:
@scaytrase That's the exact use-case I was trying to test - one of the data points is a monetary value. We could indeed represent it as integer instead, but it made sense to me to print it in decimal format given that JSON is effectively a presentation layer.

The only way I could get these tests to pass was to change the code to:

```

        $value = (float)$data / (float)$multipleOf;
        if (abs(floor($value) - $value) >= 0.00001) {
            throw KeywordMismatch::fromKeyword('multipleOf', $data, sprintf('Division by %d did not resulted in integer', $multipleOf));
        }
```

This feels like it's just kicking the problem further into the decimal wastelands though!

scaytrase:
Passing floats as a part of API can make you into different kind of troubles with intermediate gateways, since serialization\deserialization. If you really want to pass floats as a part of API I can only recommend you to cast them to strings

PR #120, opened 2021-02-27
Move psr7-cookies to dev dependencies

hansott:
I didn't find any usage in the src directory, just for tests? Cool to see my library being used in this awesome project. ❤️


scaytrase (approved):
Nice catch!

PR #135, opened 2021-07-12
(#134) Add support for comma-separated list of encoding content types

### LINKED ISSUE #134: Multipart encoding validation only checks against a single content type
According to the OpenAPI spec for the `Encoding` object's `contentType` field (this also exists in a comment in `MultipartValidator::detectEncondingContentType()`:

> The value can be a specific media type (e.g. `application/json`), a wildcard media type (e.g. `image/*`), **or a comma-separated list of the two types**

The multipart validation as it currently exists only validates against `contentType` as a single entry, failing if `contentType` is defined as a comma-separated list.

I'll submit a PR for this.

jimcottrell:
My proposal to fix #134, adding support for lists of valid encoding content types.

It should be relatively straightforward, and of course I'm open to whatever changes need to be made.

### DIFF
--- a/src/PSR7/Validators/BodyValidator/MultipartValidator.php
+++ b/src/PSR7/Validators/BodyValidator/MultipartValidator.php
@@ -28,7 +28,7 @@ use Riverline\MultiPartParser\Converters\PSR7;
 use Riverline\MultiPartParser\StreamedPart;
 use RuntimeException;
 
-use function array_diff;
+use function array_diff_assoc;
 use function array_map;
 use function array_replace;
 use function array_shift;
@@ -245,14 +245,14 @@ class MultipartValidator implements MessageValidator
         }
 
         // Any expected parameter values must also match
-        return ! array_diff($expectedNormalized, $matchNormalized);
+        return ! array_diff_assoc($expectedNormalized, $matchNormalized);
     }
 
     /**
      * Per RFC-7231 Section 3.1.1.1:
      * "The type, subtype, and parameter name tokens are case-insensitive. Parameter values might or might not be case-sensitive..."
      *
-     * Section 3.1.1.2: "A charset is identified by a case-insensitive token."
+     * And section 3.1.1.2: "A charset is identified by a case-insensitive token."
      *
      * The following are equivalent:
      *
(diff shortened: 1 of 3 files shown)


scaytrase:
Can you please provide more testcases showcasing usecases for these changes? Documentation is very unclear about how to process multiple content types. The example makes me think, that data should match one of the content-types, but I see no explicit clarification of that in the documentation.
image


https://swagger.io/specification/#encoding-object

jimcottrell:
The documentation for `contentType` is definitely a little sparse, and the example throws the curveball of also specifying `charset`, which the spec itself doesn't really mention as even being an option. I didn't really fully consider that, and this may need to be a little smarter, rather than just matching the full string value of the `Content-Type` header.

That said, I can certainly add more tests, but I'm not totally clear on what you mean by "how to process multiple content types," so I'm not sure which test cases would resolve that question. A request property can of course only have one content type, so I don't know how else it could match against the list unless it's matching against any one of the listed types.

lezhnev74:
Some extra context here:
- we have to distinguish the format of the actual `Content-Type` field (as specified by RFC 7231)
- and openapi specs as specified by openapi specification
- RFC 7231 (section [3.1.1.1](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.1)) says that media type must be a single value, however openapi allows multiple values. The only interpretation of multiple values that I can see is "any must match".
- Given the above, this PR makes sense to me

p.s. @jimcottrell I am curious, in what usecases you can use `MultipartValidator@validatePlainBodyMultipart`? Usually the message is parsed by PHP itself and popualtes $_POST variable (in which case you can't access the original http message).

jimcottrell:
@lezhnev74 For me, this issue came up in functional tests written against an API. Requests are being validated against the spec as part of the tests, so the validator sees the full request data from the requesting side.

dwilmer-karify:
This PR is exactly what I need! Does it still need the extra tests, or is there another way I can maybe help to get this merged?

jimcottrell:
I've brought the PR up to date, gone with a different approach since the list of content types seems to use the full spec from RFC 7321 (implied by the example, not explicitly stated in the OpenAPI spec), and added more thorough tests.

The content type matching could potentially be pulled out of the multipart validation, but it didn't appear that anything else looking at content types works quite the same way.

scaytrase (approved):

scaytrase:
Thanks @jimcottrell !

PR #137, opened 2021-08-02
Properly validate PUT requests

IgalOgonov:
As it stands, the request body of PUT requests is simply ignored.

This stems from the reliance on the Symfony ServerRequestInterface/getParsedBody() function, which, in case of a PUT request, will simply ignore the body. Of course, that's the intended behavior of that function, but that is not the behavior expected in the context of this validator, and in fact, can create a very serious security risk due to being relatively easy to miss.

As such, I offer a hotfix, in the form of the added function parse_raw_http_request() (taken from stackoverflow, all credit to the author, thread is linked), which parses body parameters in case of a PUT request, and using it in that case.

This is just a hotfix, as the author of that function warned it might not cover some edge cases, but in my opinion it's better than having no validation of PUT body parameters at all.

### DIFF
--- a/src/PSR7/Validators/BodyValidator/MultipartValidator.php
+++ b/src/PSR7/Validators/BodyValidator/MultipartValidator.php
@@ -231,11 +231,17 @@ class MultipartValidator implements MessageValidator
         ServerRequestInterface $message,
         Schema $schema
     ): void {
-        $body = (array) $message->getParsedBody();
 
-        $files = $this->normalizeFiles($message->getUploadedFiles());
-
-        $body = array_replace($body, $files);
+        switch($_SERVER['REQUEST_METHOD']){
+            case 'PUT':
+                $body = [];
+                $this->parse_raw_http_request($body);
+                break;
+            default:
+                $body = (array) $message->getParsedBody();
+                $files = $this->normalizeFiles($message->getUploadedFiles());
+                $body = array_replace($body, $files);
+        }
 
         $validator = new SchemaValidator($this->detectValidationStrategy($message));
         try {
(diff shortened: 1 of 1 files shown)


scaytrase (changes requested):
All tests are obviously failing, since this behaviour is not really testable and explicitly operates global state which is bad practice.

Also no new tests provided, which should describe given problem.

IgalOgonov:
> 
> 
> All tests are obviously failing, since this behaviour is not really testable and explicitly operates global state which is bad practice.
> 
> Also no new tests provided, which should describe given problem.

This is one of those commits after a sleepless night, pretty busy right now, but give me a few days and I'll see how I can properly implement this.

scaytrase:
I hope the proper way to fix this is to bring this logic into PSR-7 implementation libraries, not to the PSR-7 user libraries. Because this may lead to the problem that the PUT request content is not used in the same way as it was validated, so this will lead to errors and custom hack in this code. We should rely on standard approaches here, not the custom code.

sunkan:
This code make way to many assumptions about the environment it's running in

It assumes:
* the request body is in `php://input` that's not true if you do async development or use bref without php-fpm
* that all `PUT` request most be `multipart/form-data` witch is definitely not true
* `$_SERVER` is populated for most of my application it's not
* it ignores `getParsedBody()` for `PUT` request even thru it might not be empty

Sorry for being a bit harsh but this is just the wrong way to solve the problem your having

Now have you had a look at something like https://github.com/Kekos/parse-request-body-middleware that you could invoke before using the validator and it would handle the parsing for you?

scaytrase:
Thanks @sunkan for deep explanation! Fixing all these points will lead to fixing request object (with middleware) or request factory instead, so I think dealing with this problem is out of scope of this library

PR #145, opened 2021-10-21
Fix string-to-boolean casting

Yarre:
It is very common that boolean value casted to "0" or "1" values. And it is the industry-standard in PHP because it is how http_build_query() casts boolean. 
I faced with many such issues and have to disable validation or use artificial string in the specification when dealing with boolean flags in the GET parameters. 
So I create this small PR to fix this issue.

### DIFF
--- a/src/PSR7/Validators/SerializedParameter.php
+++ b/src/PSR7/Validators/SerializedParameter.php
@@ -125,8 +125,13 @@ final class SerializedParameter
      */
     private function castToSchemaType($value, ?string $type)
     {
-        if (($type === CebeType::BOOLEAN) && is_scalar($value) && preg_match('#^(true|false)$#i', (string) $value)) {
-            return is_string($value) ? strtolower($value) === 'true' : (bool) $value;
+        if ($type === CebeType::BOOLEAN && is_scalar($value)) {
+            if (preg_match('#^(true|false)$#i', (string)$value)) {
+                return is_string($value) ? strtolower($value) === 'true' : (bool)$value;
+            }
+            if (preg_match('#^(0|1)$#i', (string)$value)) {
+                return (string)$value === '1';
+            }
         }
 
         if (
--- a/tests/PSR7/CookieDeserializeTest.php
+++ b/tests/PSR7/CookieDeserializeTest.php
@@ -13,13 +13,28 @@ use function sprintf;
 
 final class CookieDeserializeTest extends BaseValidatorTest
 {
-    public function testItDeserializesServerRequestCookieParametersGreen(): void
+    /**
+     * @return mixed[][]
+     */
+    public function dataProviderCookiesGreen(): array
+    {
+        return [
+            ['num' , '-1.2'],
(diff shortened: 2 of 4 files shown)


lookyman:
Ping. It is quite annoying not being able to use boolean query parameters in https://github.com/OpenAPITools/openapi-generator because of this.

Yarre:
I updated tests to pass. Can you please check?

scaytrase:
Builds are failing but looks like it's not problem of the PR. I'll take a look

src/PSR7/Validators/SerializedParameter.php
```diff
@@ -125,8 +125,13 @@ private function isJsonContentType(): bool
      */
     private function castToSchemaType($value, ?string $type)
     {
-        if (($type === CebeType::BOOLEAN) && is_scalar($value) && preg_match('#^(true|false)$#i', (string) $value)) {
-            return is_string($value) ? strtolower($value) === 'true' : (bool) $value;
+        if ($type === CebeType::BOOLEAN && is_scalar($value)) {
+            if (preg_match('#^(true|false)$#i', (string)$value)) {
+                return is_string($value) ? strtolower($value) === 'true' : (bool)$value;
```
scaytrase, line 130: Looks like there is no need to return `(bool) $value here anymore, since value is either string `true` or `string` false here. I think this didn't work before either for the same reason

scaytrase:
I see a general problem with that PR, since OpenAPI 3 clearly states, that other values are not considered boolean. On the one hand one can treat this values as invalid when passed to boolean-typed header

https://swagger.io/docs/specification/data-models/data-types/

image

But raw http (headers, query args, form params are not data-typed on the other hand, so there is always string being casted. There is no real `true` or `false` value for them, only casted strings

src/PSR7/Validators/SerializedParameter.php
scaytrase, line 130: Actually this whole check looks invalid from start. I suggest to extract `if (!is_string) ` clause as first check, returning explicit `(bool) $value` as a result and treat `$value` as a string aftert (removing all redundand checks and casts)

scaytrase (approved):
I think my comments could be refactored later, LGTM 👍

scaytrase:
Thanks @Yarre, sorry for very long mering

Yarre:
Thank you! I thought it cannot be merged.

src/PSR7/Validators/SerializedParameter.php
shadowhand: Couldn't all of this be replaced with `filter_var($value, FILTER_VALIDATE_BOOLEAN)` ?

scaytrase: `filter_var` allows much more words than openapi wants

https://3v4l.org/IRcHR

shadowhand: Alright, fair enough. Just seems repetitive.

PR #168, opened 2022-05-13
Find suitable value for oneOf property in parameters. Fix #166 

### LINKED ISSUE #166: In case of specified oneOf property in parameters library does not find suitable
When validating a request using an OpenAPI specification that contains a Parameter definition as in this example:

```
        parameters:
            - in: query
              name: filter
              style: deepObject
              allowReserved: true
              schema:
                type: object
                properties:
                  filters:
                    oneOf:
                        -
                            type: object
                            required:
                              - max
                            properties:
                              max:
                                type: number
                        -
                            type: object
                            required:
                              - min
                            properties:
                              min:
                                type: number
```

and we pass correct query parameter value: ```filter[filters][min]=1``` library does not try choose suitable variant under ```oneOf``` and return error:

```
1) League\OpenAPIValidation\Tests\FromCommunity\Issue222Test::testIssue140
League\OpenAPIValidation\PSR7\Exception\Validation\InvalidQueryArgs: Value "{
    "filters": {
        "min": "1"
    }
}" for argument "filter" is invalid for Request [get /list]

```

sloory:


sloory:
this PR fix #166 166

src/PSR7/Validators/SerializedParameter.php
```diff
@@ -181,6 +183,14 @@ protected function convertToSerializationStyle($value, ?CebeSchema $schema)
         if ($schema && $this->style === self::STYLE_DEEP_OBJECT) {
             foreach ($value as $key => &$val) {
                 $childSchema = $this->getChildSchema($schema, (string) $key);
+
+                if (isset($childSchema->oneOf)) {
```
scaytrase, line 187: Looks like in this case we should immediately stop validation if we have filled oneOf and none of the schemas matched in `findSuitableOneOf`.

tests/FromCommunity/Issue166Test.php
```diff
@@ -0,0 +1,72 @@
...
+        $query       = null;
+        parse_str($queryString, $query);
+
+        $psrRequest = (new ServerRequest('get', 'http://localhost:8000/api/list'))
+            ->withHeader('Content-Type', 'application/json')
+            ->withQueryParams($query);
+
+        try {
+            $validator->validate($psrRequest);
+            $this->assertTrue(true);
```
scaytrase, line 67: this is not proper way how to deal with exceptions and zero assertions. proper way would be (without any try..catch)

```
$validator->validate($psrRequest);
$this->addToAssrtionsCount(1);
```

any exception thrown here should break the test and if `validate` throws one `assertTrue(true)` is not called anyway

tests/FromCommunity/Issue166Test.php
```diff
@@ -0,0 +1,72 @@
...
+use League\OpenAPIValidation\Tests\PSR7\BaseValidatorTest;
+
+use function parse_str;
+
+/**
+ * @see https://github.com/thephpleague/openapi-psr7-validator/issues/166
+ */
+final class Issue166Test extends BaseValidatorTest
+{
+    public function testIssue166(): void
```
scaytrase, line 19: some negative cases required

scaytrase (changes requested):
Needs some fixing for negative cases

PR #171, opened 2022-07-21
fix(floating-point): use bcmath extension instead of simple division

Ni-vek:


Ni-vek:
#172

scaytrase (approved):

scaytrase:
Sorry for ling delay. Good one, thanks @Ni-vek !

PR #175, opened 2022-08-05
Prevent ErrorException when value is not iterable

rogervila:
Hello,

When `$value` is a string, an `ErrorException` is thrown with the following message:

```sh
foreach() argument must be of type array|object, string given
```

This PR prevents that exception by checking if `$value` is an array, as expected.

### DIFF
--- a/src/PSR7/Validators/SerializedParameter.php
+++ b/src/PSR7/Validators/SerializedParameter.php
@@ -171,8 +171,8 @@ final class SerializedParameter
                 $value = explode(self::STYLE_DELIMITER_MAP[$this->style], $value);
             }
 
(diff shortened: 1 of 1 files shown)


rogervila:
Hi @scaytrase 

I am thinking about using `is_iterable` instead of `is_array`. WDYT?

scaytrase:
Sounds great

scaytrase (approved):

rogervila:
@scaytrase done :) 

Thank you for the review!

scaytrase (approved):

scaytrase:
Thanks @rogervila !

PR #177, opened 2022-09-02
Fix missing ExceptionInterface

rogervila:
Hi,

`\Respect\Validation\Exceptions\ExceptionInterface` does not exist anymore, and `Respect\Validation\Exceptions\Exception` extends from `\Throwable` without adding any additional value.

This PR removes the unexisting interface usage and simplifies exception catching.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -5,5 +5,4 @@ parameters:
         - tests
     treatPhpDocTypesAsCertain: false
     ignoreErrors:
-        - '#Caught class Respect\\Validation\\Exceptions\\ExceptionInterface not found.#'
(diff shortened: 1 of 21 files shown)


scaytrase:
Sorry for a very long responses. AFAIK these exceptions are for `^1.1.3` version dependency stated in `composer.json`. We need to ensure that these interfaces are non-existent event in lower deps versions and\or update dependencies properly.

rogervila:
Thank you @scaytrase 

Since `\Throwable` can catch any exception, this PR should not be a breaking change. 

Let me know if there is any issue related to this change so I can fix the PR.

scaytrase (approved):

scaytrase:
Thanks @rogervila !

PR #180, opened 2022-12-06
Fix tests and phpstan

OliverProgrammer:


OliverProgrammer:
Sry a mistake please ignore

Issue #181, opened 2022-12-07
Error validating query arguments of type array


mam08ixo:
There are several ways to express arrays in query strings:

The array

```
"list": [
    "foo",
    "bar"
]
```

may be serialized into `list[]=foo&list[]=bar` or `list=foo&list=bar`.

Depending on the server backend, the _latter_ variant may be required. When using this format though, request validation breaks:

https://github.com/thephpleague/openapi-psr7-validator/blob/5f98f98abf37f4533473699ef2ff2b4dc9b8d52e/src/PSR7/Validators/SerializedParameter.php#L174-L176

**`Invalid argument supplied for foreach()`**

The issue is caused by the usage of the `parse_str` function, which can only handle PHP-style query args and turns multiple arguments with the same name into a string with the last value ("bar"):

https://github.com/thephpleague/openapi-psr7-validator/blob/5f98f98abf37f4533473699ef2ff2b4dc9b8d52e/src/PSR7/Validators/QueryArgumentsValidator.php#L74

This drawback is also mentioned in the [user contributed notes](https://www.php.net/manual/en/function.parse-str.php#76792) of the `parse_str` documentation.

The objective would be to properly parse all array query string formats that are mentioned in the [OpenAPI Guide](https://swagger.io/docs/specification/describing-parameters/#query-parameters).

mam08ixo:
[`\GuzzleHttp\Psr7\Query::parse`](https://github.com/guzzle/psr7/blob/2.4.3/src/Query.php#L10-L20) handles my use case pretty well, but, as noted there, not the PHP style array.

PR #162, opened 2022-04-06
Fix #147: multipart/form-data validation

vovochka404:
Fix issue #147

PR #182, opened 2022-12-17
acce

tyamahori:
# ref #181

Hi.

### DIFF
--- a/src/PSR7/Validators/QueryArgumentsValidator.php
+++ b/src/PSR7/Validators/QueryArgumentsValidator.php
@@ -15,7 +15,9 @@ use Psr\Http\Message\MessageInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
-use function parse_str;
+use function array_filter;
+use function explode;
+use function urldecode;
 
 /**
  * @see https://swagger.io/docs/specification/describing-parameters/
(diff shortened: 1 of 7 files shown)


katsumi-kato-LITALICO (approved):
sorry.

scaytrase:
@tyamahori Hi, same question as here https://github.com/thephpleague/openapi-psr7-validator/pull/162#discussion_r1088794714 

Isn't this notation valid only in deepObject serialization form ? I think we need to check this explicitly here

tyamahori:
@scaytrase 
Hi, thank you for your comment.

This notation is valid if a query string is `paramName[]=value1&paramName[]=value2&.... `


> deepObject – simple non-nested objects are serialized as paramName[prop1]=value1&paramName[prop2]=value2&.... The behavior for nested objects and arrays is undefined.

Do you mean, you think `paramName[prop1]=value1&paramName[prop2]=value2&....` should be checked ?

scaytrase:
> Do you mean, you think paramName[prop1]=value1&paramName[prop2]=value2&.... should be checked ?

no, I mean that deafult serealization method is `form + explode` which clearly shows the example for array values

image

> `/users?id=3&id=4&id=5`

So if you want to provider query args in another format  - you should specify another serialization style in your schema. From given styles I see `deepObject` as the most suitable one here

https://swagger.io/docs/specification/serialization/

They specify `n\a` for array example for deepObject, but I think `prop[]=val1&prop[]=val2` notation is OK here

scaytrase (changes requested):
Spec for test doesn't match test data

PR #183, opened 2023-01-03
Access operation security element directly, avoiding getSerializableData

jimcottrell:
With a relatively full-featured spec, `SpecFinder::findSecuritySpecs()` is pretty slow because of the call to `getSerializableData`. However, it's not necessary to serialize the `Operation` element to see if `security` is defined, as the property can be accessed directly and the getter will do the right thing.


scaytrase:
Good one. `security` property was made accessible indeed. Thanks @jimcottrell !

scaytrase (approved):

PR #187, opened 2023-01-29
Ensure additionalProperty uses correct schema.

fiasco:
I encountered this issue when attempting to validate data against this schema:

```yaml
EnvironmentFlags:
      type: object
      description: An array of various flags that indicate functionality associated with the environment.
      additionalProperties:
        type: object
        properties:
          cde:
            type: boolean
            description: Determines if the environment is on-demand.
```

The `Properties::validate()` method currently attempts to validate `additionalProperties` by their parent definition which results in an error which validating valid data.

### DIFF
--- a/src/Schema/Keywords/Properties.php
+++ b/src/Schema/Keywords/Properties.php
@@ -108,7 +108,7 @@ class Properties extends BaseKeyword
             // if not covered by "properties"
             $schemaValidator->validate(
                 $data[$propName],
-                $additionalProperties,
+                $additionalProperties->properties[$propName],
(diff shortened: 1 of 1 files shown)


fiasco:
I've added a unit test for this scenario and updated the PR to support both existing and new scenarios.

scaytrase (changes requested):
I'm not sure I'm getting this correct. I see incorrect match for spec and test data in your UT.

if  you have additional properties being specd by `object: props: flag: bool` than your test data should look like

```php
$data = ['someUnknownProperty' => ['flag' => true]];
```

For current format `flag` should be known property of main object, not for additionalProperties

PR #191, opened 2023-02-08
Handle path parameters for arrays correctly

tjcambio:
The parameters for serialized arrays was using the same deserialization check.
There are different formats for header, query and path. Split up the validation of arrays into two different sets to allow different validation for path and header/query parameters.

The changes will validate path parameters for arrays but not objects which is still not working.

Because of objects within path parameter is probably the most unused things in the specs these have been left out intentionally.


scaytrase:
Thanks @tjcambio !

PR #192, opened 2023-02-16
Allow specifying encoding for optional parts of multipart data

jimcottrell:
Currently, any part of multipart data with an `encoding` provided in the spec will throw a validation error if it's not present in the body. As far as I know, providing an `encoding` value is not supposed to imply `required` for the part, so I've simply removed the check throwing an exception if a part doesn't exist when checking for a valid encoding.

I've expanded the test spec for this case and for good measure, I've added a test showing that the schema validation takes place, though I think other multipart tests probably do cover that.

@lezhnev74 Do you know if I'm missing something on this?


scaytrase (approved):
LGTM 👍

scaytrase:
Oh there is online conflict resolution tool now, nice

scaytrase:
Thanks @jimcottrell !

Issue #172, opened 2022-07-21
FLoating point issue with multipleOf rule


Ni-vek:
Hello,

These values fail with multipleOf rule, due to floating point precision

- value to validate : 0.94
- multiple of : 0.01

A possible solutioin is to use bcmath fiunction

scaytrase:
ehm, @bizurkur is `8` is multipleOf `1.998` ? but `round(8 / 1.998, 2)` is precisely `4`

https://3v4l.org/525Cf

scaytrase:
> Using that logic, the current "fix" is also just as broken: https://3v4l.org/JYC3W

Wow, that's good catch. Actually I'll consider revert this then, for me ext dependency is also looks bad, but at least I thought it brings some profit

PR #194, opened 2023-04-03
Use epsilon to compare floating points

bizurkur:
This is a alternative to https://github.com/thephpleague/openapi-psr7-validator/pull/171 that doesn't require everyone to install bcmath to keep using this library.

Precision of 8 should be "close enough" to consider it equal. That's one 10 millionth of a fraction off... which is an incredibly small amount. `1 / 3 = 0.33333333` therefore `1 / 0.33333333 = 3` because it is "close enough."

Issue #172

### DIFF
--- a/tests/Schema/Keywords/MultipleOfTest.php
+++ b/tests/Schema/Keywords/MultipleOfTest.php
@@ -21,6 +21,7 @@ final class MultipleOfTest extends SchemaValidatorTest
             [10, .5],
             [9.9, .3],
             [.94, .01],
+            [1, .3333333333],
         ];
     }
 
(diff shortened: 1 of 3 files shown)


tests/Schema/Keywords/MultipleOfTest.php
```diff
@@ -21,6 +21,7 @@ public function validDatasets(): array
             [10, .5],
             [9.9, .3],
             [.94, .01],
+            [1, .3333333333],
```
scaytrase, line 24: Despite the fact it is formally correct according to current epsilon I'd rather avoid fix this as a valid behaviour

scaytrase (approved):
One small change suggested, but looks great!

tests/Schema/Keywords/MultipleOfTest.php
scaytrase: yep, just drop it

PR #185, opened 2023-01-26
Use devizzent/cebe-php-openapi to be compatible with OpenApi 3.1

DEVizzent:

Issue #163, opened 2022-04-08
OpenAPI v3.1 Support


philsturgeon:
Hello everyone! As the maintainer of [openapi.tools](https://openapi.tools) I'm interested in tracking progress towards supporting OpenAPI v3.1. 

It's got a bunch of amazing changes in it, solving problems like the the [JSON Schema <!=> OpenAPI Schema Object divergence](https://apisyouwonthate.com/blog/openapi-json-schema-divergence), and fixes some other inconsistencies and duplicate ways of doing things. It's the best version and everyone should be using it, and I see that some folks are starting to use it as they're getting problems #148. We need the tooling to catch up, and its been over a year now. Just in case folks didn't notice, I'm here to give a friendly prod! 

Here are a few articles showing off the differences between OpenAPI v3.0 and v3.1.

https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0
https://www.apimatic.io/blog/2021/09/migrating-to-and-from-openapi-3-1/ 
https://nordicapis.com/whats-new-in-openapi-3-1-0/

If you say you'll be working on it I can update [openapi.tools](https://openapi.tools) to reflect that. 

Let me know if you have any questions, or if you need any help with the upgrade process.

cebe:
Support of OpenAPI 3.1 in this library heavily depends on `cebe/php-openapi`. Progress of that is tracked here: https://github.com/cebe/php-openapi/issues/101

cebe:
> If you say you'll be working on it I can update [openapi.tools](https://openapi.tools) to reflect that.

@philsturgeon please update openapi.tools status for `thephpleague/openapi-psr7-validator` and `cebe/php-openapi` to WIP.

philsturgeon:
@cebe done, thanks for the update!

tomasbreffitt:
With #185 merged, can we get a tagged release?

scaytrase:
released as 0.19

pieterocp:
We can probably close this now that #185 has been merged and released.

Tim-Haboldt:
Am I correct in understanding that OpenAPI v3.1 is now largely supported, with the exception of the "unevaluatedProperties" keyword?

philsturgeon:
@Tim-Haboldt depends very much of what you mean by that. Any tool using a compliant JSON Schema tool will support `unevaluatedProperties` through validation or similar, but a lot of documentation tools might not know how to show that (or various other new things).

[...]

PR #196, opened 2023-05-03
Update README.md

MelchiorKokernoot:
Given this PR has been merged, I think 3.1.x is supported now? The readme still said 3.0.x

https://github.com/thephpleague/openapi-psr7-validator/issues/163
https://github.com/thephpleague/openapi-psr7-validator/pull/185


pieterocp:
Probably want to make it clear it supports both 3.0.x and 3.1.0 instead of less than 3.1.X since I don't think that version 2 is supported by this tool? (But I might be mistaken)

MelchiorKokernoot:
Sure thing, changed it, never saw your comment.

PR #203, opened 2023-06-13
Fix Invalid "items MUST be present if the type is array" with OpenAPI 3.1

### LINKED ISSUE #202: Invalid "items MUST be present if the type is array" with OpenAPI 3.1
The error "`items MUST be present if the type is array`" is returned since openApi 3.1, for nullable arrays, as the type is now an array.

```php
if (! isset($this->parentSchema->type) || ($this->parentSchema->type !== 'array')) {
            throw new InvalidSchema(sprintf('items MUST be present if the type is array'));
        }
```

The value of `$this->parentSchema->type` can be an array with : 
- "array"
- "null"

momala454:
Fixes https://github.com/thephpleague/openapi-psr7-validator/issues/202

### DIFF
--- a/src/Schema/Keywords/Items.php
+++ b/src/Schema/Keywords/Items.php
@@ -46,7 +46,7 @@ class Items extends BaseKeyword
             throw InvalidSchema::becauseDefensiveSchemaValidationFailed($e);
         }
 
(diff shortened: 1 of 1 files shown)


bambamboole:
Thats my fix :) How can we get this in ?

PR #204, opened 2023-07-21
Use 'object given' in TypeMismatch msg when expecting array

### LINKED ISSUE #25: Distinguish between assoc arrays and objects in validation messages

dhaeckel:
resolves #25


drealecs (approved):
This looks good as I stumbled on the same problem with the not clear message:
`"Value expected to be 'array', 'array' given."`.

Can we have it merged?

PR #205, opened 2023-08-10
Updated dependencies: psr/http-message 2.x & league/uri 7.x

deegital:
- Support for psr/http-message 2.x
v2.x added return types to the interfaces, should be fine - tests running

- Support for league/uri 7.x
This package is only used within `\League\OpenAPIValidation\Schema\TypeFormats\StringURI` to check if an url is parsable:
```php
try {
    UriString::parse($value);

    return true;
} catch (SyntaxError $error) {
    return false;
}
```


MaSpeng:
An Update for `respect/stringifier` might also be good, and could also be done in a separate PR.

deegital:
@MaSpeng I've just taken a look at it, but noticed it's not a explicit dependency of this package, but an implicit one via `respect/validation`.

The latest version of that package is already supported by the current required version:
 `"respect/validation": "^1.1.3 || ^2.0",`

Requests for bumping the stringy version then need to be added here: https://github.com/Respect/Validation
Hope that helps :)

schodemeiss:
Any chance of progressing this PR at all?

I have a dependency on `league/uri ^7.0`, but this package is preventing that upgrade unfortunately :-(.

deegital:
@schodemeiss feel free to use my branch in the meantime by adding the repository to your composer file
```
"repositories": [
    {
        "type": "vcs",
        "url": "https://github.com/deegital/openapi-psr7-validator"
    }
],
```

and this version of the package to require/require-dev:
```
"league/openapi-psr7-validator": "dev-dependency-updates",
```

I will keep it at least until the PR is merged or it is added to the lib by other commits

deegital:
@scaytrase sorry for tagging, but I've noticed you did the last approval of an MR for this repo. What's missing to move on with this MR?

PR #215, opened 2024-01-02
Expand CI coverage and tighten execution

KentarouTakeda:
## Overview

* Added 8.2 and 8.3 to the php version to be tested in CI.
* Fixed test failure on php-8.1 and later by using phpunit according to php version and officially added to CI.

## Detail

A detailed step-by-step explanation of all the fixes made in this pull request.

1. https://github.com/thephpleague/openapi-psr7-validator/commit/3472e1e9fb15b675318693470bcf5377309242e1 : Added 8.1, 8.2, 8.3 to the php version to be tested.
   * All tests now fail on php 8.1 and later, which were previously marked as `experimental`. [Here](https://github.com/KentarouTakeda/thephpleague-openapi-psr7-validator/actions/runs/7382957696) are the results.
   * This is because `phpunit:^7` installed with `--prefer-lowest` does not support php-8.1 or later.
2. https://github.com/thephpleague/openapi-psr7-validator/commit/a0485215eec9e58ff38607c00a448e6858dbfebf : Use different versions of phpunit depending on the php version to be tested.
   * With this fix, all tests that previously failed now pass.
   * So far, most of the issues have been resolved, but the CI results are showing many warning messages below. Here are the [results]((https://github.com/KentarouTakeda/thephpleague-openapi-psr7-validator/actions/runs/7382965349).
     > The following actions uses node12 which is deprecated and will be forced to run on node16: shivammathur/setup-php@2.9.0, actions/checkout@v2.
3. https://github.com/thephpleague/openapi-psr7-validator/commit/954ec86d51d17318fcdedec85e5f86104afde4a4 : Use the latest version of `setup-php`.
    * All CI warning messages have disappeared. [Here](https://github.com/KentarouTakeda/thephpleague-openapi-psr7-validator/actions/runs/7382971012) are the results.

## Motivation

According to [Packagist usage statistics](https://packagist.org/packages/league/openapi-psr7-validator/php-stats#0.21), more than 90% of the versions of php currently installed with this library are 8.1 or later. Testing on real versions will improve reliability.

PHP-Versions-statistics

These tests may also be required when changing the version of dependent libraries. This is because the updated version is likely to use a newer PHP version.

For example, the following pull request also suggests a broader version dependency:

* [Add support to psr/http-message 2.0 #213](https://github.com/thephpleague/openapi-psr7-validator/pull/213)


KentarouTakeda:
Adjusted the way dependencies are specified to pass the `Scrutinizer` CI. Ready for review.

PR #221, opened 2024-03-15
Allow null as standalone type

JurianArie:
The validator throws an exception with the following (API enforces that only one of the fields is filled):

Data passed:
```php
[
    'department_id' => null,
    'team_id' => 1,
]
```

Schema:
```yaml
- oneOf:
    - type: object
      properties:
        department_id:
          type: integer
        team_id:
          type: 'null'
      required:
        - department_id
    - type: object
      properties:
        department_id:
          type: 'null'
        team_id:
          type: integer
      required:
        - team_id
```

With the proposed change the validator would not throw an exception any more.

### DIFF
--- a/src/Schema/Keywords/Nullable.php
+++ b/src/Schema/Keywords/Nullable.php
@@ -27,6 +27,10 @@ class Nullable extends BaseKeyword
 
     public function nullableByType(): bool
     {
-        return ! is_string($this->parentSchema->type) && in_array('null', $this->parentSchema->type);
+        if (is_string($this->parentSchema->type)) {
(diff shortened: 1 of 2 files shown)


leNEKO:
Hi, it seems that there is not a such `null` type in the openapi specification (see https://swagger.io/docs/specification/data-models/data-types/#null).

JurianArie:
@leNEKO openapi 3.1 does have a "null" type: https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0 under the section "Swap nullable for type arrays"

PR #228, opened 2024-08-21
Updated library league/uri

lucapacchiarotta:
Current constraint on library version `"league/uri": "^6.3 || ^7.0"`, make not possible to install the most recent library `league/uri-interfaces ^7.4`, that is required in some important packages, as the amphp/http-client.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         }
     },
     "require": {
-        "php": ">=7.2",
+        "php": ">=7.2 || ^8.1",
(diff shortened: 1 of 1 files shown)


garak:
You can check on https://jubianchi.github.io/semver-check/#/^6.3%20||%20^7.0/7.4.0 that the current constraint already allows the 7.4 version of `league/uri`, so your proposed change doesn't make much sense.

PR #239, opened 2024-12-16
Feat: make the keyword compliant with OpenApi 3.1

p-pichet:
When usign to validate a schema with a nullable array 
```php
class foo
{ 
    /** @var ?array $bar */
   protetected ?array $bar;
   }
   ```
   The type is an array with the values `array' and `null` on the OpenApi 3.1
   
   With this fix, we check if the type is string or an array.


p-pichet:
@scaytrase I see you are the last that merge on this project.  
Can we see if it's good ?

PR #243, opened 2025-04-16
Bug Fix: Incorrect Static Paths Prioritising

halidovz:
**Problem:**  
When OpenAPI `servers` are defined and there are multiple paths with the same number of segments—some dynamic and some static—the validator may incorrectly match both paths.

**Example OpenAPI definition:**
```yaml
servers:
  - url: https://localhost/v1

paths:
  get /products/{product}/images/{image}
  get /products/{product}/images/thumbnails
```

**Incoming request:**
```
GET /v1/products/10/images/thumbnails
```

**Actual behavior:**  
Both paths `/products/{product}/images/{image}` and `/products/{product}/images/thumbnails` are matched.

**Expected behavior:**  
Only the more specific, static path `/products/{product}/images/thumbnails` should match.

**Cause:**  
The current path matching logic compares only the relative path without properly including the server prefix (e.g., `/v1`). This leads to ambiguous or incorrect matching when evaluating path candidates of equal length.

**Fix:**  
Ensure that the matching logic compares full paths—including the server prefix—when determining the best match.

### DIFF
--- a/src/PSR7/OperationAddress.php
+++ b/src/PSR7/OperationAddress.php
@@ -28,6 +28,8 @@ class OperationAddress
     protected $method;
     /** @var string */
     protected $path;
+    /** @var string */
+    protected $fullPath;
 
     public function __construct(string $path, string $method)
     {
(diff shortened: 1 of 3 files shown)


halidovz:
Hey @lezhnev74 , could you please review this when you get a chance?

PR #251, opened 2025-06-09
Upgrade ci php versions

pieterocp:
Upgrade to use the latest php versions, so we can be more sure any fixes should work on a nice range of php versions.

Seem to be all passing on my fork, upgraded `shivammathur/setup-php` too while at it since presumably there's some performance gains hidden in there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,14 +63,14 @@ jobs:
         run: ./vendor/bin/phpunit -c phpunit.xml
 
   cs:
-    name: Codestyle check on PHP 8.3
+    name: Codestyle check on PHP 7.2
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -51,14 +63,14 @@ jobs:
         run: ./vendor/bin/phpunit -c phpunit.xml
 
   cs:
-    name: Codestyle check on PHP 7.2
+    name: Codestyle check on PHP 8.3
```
pieterocp: This is a very opinionated take, really it makes sense to continue development, linting + additional checks in the lowest supported version, then polyfill + run tests upwards, ergo removing this now.

pieterocp:
Since #267 has been merged in this will be closed as this covers it well.

PR #267, opened 2026-03-03
Update ci to modern versions

shadowhand:
Refs #251


shadowhand:
There's nothing I can do about Scrutinizer. IMO, that should be dropped entirely.

